PR #22, opened 2020-07-18
Add list of event keys to allow for Siddhi PEs with multiple streams

amkcpu:
### Purpose
1. Currently, there is only one list of event keys (attributes), which gets expanded with the keys from all streams: `sortedEventKeys`. When connecting multiple streams to a Siddhi processing element, this leads to a `NullPointerException` and duplicate keys in the stream definition clause. This PR adds a HashMap containing a separate list of event keys for each stream.
2. If both connected streams had the same attribute name, a Siddhi exception was raised. This PR changes attribute naming to that the second stream is prefixed "s1", the third "s2", etc. So if using the integrated Flow Rate 1 and Flow Rate 2 adapters, this would mean the mass_flow attribute is called `s0mass_flow` for the first and `s1mass_flow` for the second stream.

### Approach
1. Adds the `listOfEventKeys` HashMap. Keys are the stream names, values are the lists containing the respective event key. This is somewhat similar to the already existing `siddhiInputHandlers` Map, just with a different purpose.
2. Changes `registerEventTypeIfNotExists()` to put new event keys there when invocating the processing element.
3. Changes `onEvent()` and `toObjArr()` to use the HashMap when receiving new events.
4. Change attribute naming in `registerEventTypeIfNotExists()` by counting the number of elements in `listOfEventKeys` and prefixing accordingly.

### Remarks
Jira issue: None

### DIFF
--- a/streampipes-wrapper-siddhi/src/main/java/org/apache/streampipes/wrapper/siddhi/engine/SiddhiEventEngine.java
+++ b/streampipes-wrapper-siddhi/src/main/java/org/apache/streampipes/wrapper/siddhi/engine/SiddhiEventEngine.java
@@ -294,9 +294,12 @@ public abstract class SiddhiEventEngine<B extends EventProcessorBindingParams> i
     return selectString.toString();
   }
 
-  /*
   public void setSortedEventKeys(List<String> sortedEventKeys) {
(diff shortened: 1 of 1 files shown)


wipatrick:
I tested this PR and it works as described. Thanks for the contribution @amkcpu

PR #24, opened 2020-07-20
[STREAMPIPES-172] Set default secondary text color

ajaybhat:
### Purpose

Change the default secondary text color so that it's visible from the grid.

### Remarks

Fixes: STREAMPIPES-172


wipatrick:
Hi @ajaybhat thanks for the PR :) works perfectly.

PR #25, opened 2020-07-20
[STREAMPIPES-173] Add env vars to pom.xml

ajaybhat:
### Purpose

Add all env vars from `development/env` to `pom.xml` so that Springboot can be run directly from the command line, using `mvn spring-boot:run`

### Remarks

Fixes: STREAMPIPES-173

### DIFF
--- a/streampipes-backend/pom.xml
+++ b/streampipes-backend/pom.xml
@@ -83,8 +83,24 @@
                         <goals>
                             <goal>repackage</goal>
                         </goals>
+                        <configuration>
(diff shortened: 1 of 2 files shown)


wipatrick:
Hi @ajaybhat this looks really nice. I updated the env file in the backend accordingly since we do not need Kafka Rest anymore. Could you update the entries in the pom.xml accordingly?

wipatrick:
hi @ajaybhat finally got to test your PR - thx again.

PR #32, opened 2021-01-28
[WIP] Added functionality for (partial/selective) pipeline element migration

daniel-gomm:
### Purpose

Added the functionality for partial migration, meaning the migration of single pipeline elements, while the corresponding pipeline continues running.
Allows for alterations in the pipeline deployment, without requiring stopping the pipeline first, e.g., desirable if pipeline should continue running while a node undergoes maintenance.
### Approach

Selectively start and stop specific pipeline elements (PEs) and datastream relays. With a focus on performance, the new PE is invoked first, then the event routing is adjusted accordingly before the original PE is detached.
Additionally, new endpoints for stopping and starting relays of PEs were added to the node controller.
### Samples

-
### Remarks

-


wipatrick:
Thanks for the PR and the work on the edge extension topics @daniel-gomm! I'll take a look and test it.

wipatrick:
I tested your PR and it works as aspected! Thanks again @daniel-gomm

PR #33, opened 2021-03-23
[STREAMPIPES-321] Updated flinkruntime jobstatus checks to check actual running status

mfaizanse:
### Purpose
If the flink job is not found, then it means that it is cancelled or failed. So streampipes should continue to stop this PE from its in-memory storage and mark the whole pipeline as Stopped.

### Approach
Changed in FlinkRuntime.java. Modified method discardRuntime() to also check if the job is present and its state is running then cancel the job, else if job is present but bot running then don't throw exception. 

Also change method getJobStatus() method to return job with any status.

### Remarks
Fixes: https://issues.apache.org/jira/projects/STREAMPIPES/issues/STREAMPIPES-321


wipatrick:
Hi @mfaizanse, thanks again for your PR. I just checked out your PR and it works as aspected.

PR #34, opened 2021-04-02
[STREAMPIPES-323] Implemented Merge by Schema Inbox.

Samarth08:
### Approach

Fixes: https://issues.apache.org/jira/browse/STREAMPIPES-323

### DIFF
--- a/streampipes-model/src/main/java/org/apache/streampipes/model/schema/EventProperty.java
+++ b/streampipes-model/src/main/java/org/apache/streampipes/model/schema/EventProperty.java
@@ -21,6 +21,7 @@ package org.apache.streampipes.model.schema;
 import com.fasterxml.jackson.annotation.JsonSubTypes;
 import io.fogsy.empire.annotations.RdfProperty;
 import io.fogsy.empire.annotations.RdfsClass;
+import org.apache.commons.collections.ListUtils;
 import org.apache.streampipes.model.base.UnnamedStreamPipesEntity;
 import org.apache.streampipes.model.quality.EventPropertyQualityDefinition;
 import org.apache.streampipes.model.quality.EventPropertyQualityRequirement;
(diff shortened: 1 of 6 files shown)


streampipes-model/src/main/java/org/apache/streampipes/model/schema/EventProperty.java
```diff
@@ -85,6 +86,48 @@
 
   private String runtimeId;
 
+  private boolean equateLists(List<?> list1, List<?> list2) {
```
grainier, line 89: Why don't we use `list1.equals(list2)`?

Samarth08: I have used ListUtils.isEqualList for comparing the lists now(requires extra dependency). Since we are already overriding equals everywhere for this story.
Is this fine?Or you want me to sort the list for which we have to write the comparator function and then do list1.equals(list2).

streampipes-model/src/main/java/org/apache/streampipes/model/schema/EventProperty.java
```diff
@@ -30,9 +31,7 @@
 
 import javax.persistence.*;
 import java.net.URI;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.UUID;
+import java.util.*;
```
grainier, line 34: Shall we use single imports, instead of `import *`

Samarth08: I have removed java.util.* and added single imports.

bossenti:
@Samarth08, could you please provide your commit messages in the style of "[STREAMPIPES-XXX] commit message"?
That would be great! 
Besides that, this looks good to me.

bossenti (changes requested):
reformat commit messages

Samarth08:
@bossenti I have addressed the review comments. Please have a look.

bossenti (approved):

PR #42, opened 2021-05-04
[STREAMPIPES-351] Remove invalid dependency entry to avoid start-up failure

mohanvive:
### Purpose
Fixes the https://issues.apache.org/jira/browse/STREAMPIPES-351


mohanvive:
@grainier @wipatrick @dominikriemer

dominikriemer:
Thanks @mohanvive !

PR #48, opened 2021-08-09
flink 1.13 rest port config & Keep kafka consumer group for Pipeline action and sepas

yangdh1225:
### Purpose

1. custom flink rest port from configuration for version 1.13
2. Make kafka consumer group id fixed for pipeline action and sepas

### Remarks

Fixes: <Jira issue link>

### DIFF
--- a/streampipes-pipeline-management/src/main/java/org/apache/streampipes/manager/execution/http/PipelineExecutor.java
+++ b/streampipes-pipeline-management/src/main/java/org/apache/streampipes/manager/execution/http/PipelineExecutor.java
@@ -19,6 +19,7 @@
 package org.apache.streampipes.manager.execution.http;
 
 import org.apache.streampipes.commons.MD5;
+import org.apache.streampipes.commons.Utils;
 import org.apache.streampipes.manager.execution.status.PipelineStatusManager;
 import org.apache.streampipes.manager.secret.SecretProvider;
 import org.apache.streampipes.manager.util.TemporaryGraphStorage;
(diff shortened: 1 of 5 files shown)


streampipes-pipeline-management/src/main/java/org/apache/streampipes/manager/execution/http/PipelineExecutor.java
```diff
@@ -107,7 +108,7 @@ private void updateGroupIds(InvocableStreamPipesEntity entity) {
             .filter(is -> is.getEventGrounding().getTransportProtocol() instanceof KafkaTransportProtocol)
             .map(is -> is.getEventGrounding().getTransportProtocol())
             .map(KafkaTransportProtocol.class::cast)
-            .forEach(tp -> tp.setGroupId(UUID.randomUUID().toString()));
+            .forEach(tp -> tp.setGroupId(pipeline.getName() + "-" + MD5.crypt(tp.getElementId())));
```
dominikriemer, line 111: Can we use pipeline.getId() instead of the name to avoid having blank spaces from the user-defined pipeline name in the group Id?

dominikriemer (commented):
Hi @yangdh1225 
thanks for the PR!
Is there a reason why we use the pipelineName instead of the ID for generating the groupId?

Dominik

streampipes-pipeline-management/src/main/java/org/apache/streampipes/manager/execution/http/PipelineExecutor.java
yangdh1225: The pipeline name is used instead of id, just to make it easier to identify in Kafka monitoring. Certainly if the UI interface does not check special characters, such as spaces, surely we should use id to generate group id.

yangdh1225:
> Hi @yangdh1225
> thanks for the PR!
> Is there a reason why we use the pipelineName instead of the ID for generating the groupId?
> 
> Dominik

Thanks for comment.
I have submitted a commit to replace special characters in pipeline name.

dominikriemer:
Thanks!

PR #61, opened 2021-11-23
Added Seperate textFilter1 test

Hrushi20:
I created a new folder filters-jvm in pipelineElements. Moved the test of textfilter1 to it's respective file.


Hrushi20:
This is the desired change? I am still confused on what needed to be implemented.

dominikriemer:
Hi,
the idea of the test framework is that only three configuration files are needed for any pipeline element to test.
Have a look at the folder [1]. There is a JSON file and two CSV files. The first CSV file (named input.csv by convention) specifies the input events that are used for testing the processor.  The second CSV file (expected.csv) specifies the events that are expected to be sent by the processor. The third file (description.json) specifies the configuration of the processor (what you'd normally do in the UI). The "selector" prefix matches the internal name of the static property configuration, and the value corresponds to the value that should be applied to the config.

So in order to create a new pipeline element test, you need to create a new directory similar to textFilter1 and provide these three files. The only thing left is to register the test in the framework. You don't need to copy the code as in your PR, the only change required is to add the directory of the new test to the array in SinglePipelineElementTest.ts, line 25. The test framework will pick up the test, read the configuration from your created folder and execute the test.

Hope this helps!


[1] https://github.com/apache/incubator-streampipes/tree/dev/ui/cypress/fixtures/pipelineElement/filters-jvm/textFilter1

Hrushi20:
Oh! Thank you very much. I will create another folder named textFilter2 containing more tests related to textFilter and add it to line 25 of SinglePipelineElementTest.ts

PR #63, opened 2021-12-04
[STREAMPIPES-479] Add spring-context-support dependency

obermeier:
Add 
        <dependency>
            <groupId>org.springframework</groupId>
            <artifactId>spring-context-support</artifactId>
        </dependency>

to archetype


obermeier:
Not required anymore. Alternative #64 is used

PR #71, opened 2021-12-30
[STREAMPIPES-491] - Webpack 5 issues

hutcheb:
### Purpose
I ran into some issues that seem to be related to webpack 5. 
1) When running ng build without the --configuration production option, it complained about a missing assert package
2) When running the ui using npm start, and opening in the browser. An error message is generated complaining about not being able to find the process variable.

### Approach
1) I have added the assert package to packes.json as a dev dependancy.
2) I have added the provide plugin to the webpack configuration to include 'process/browser'


dominikriemer:
Great, looks good to me!
Does the ng build command also execute the prebuild script? Maybe we could also add a script "build-dev" to the npm scripts to run a non-production build.

hutcheb:
ngx-color-picker and ng-simple-slideshow still need to be updated

dominikriemer:
Awesome, thanks and happy new year :-)

hutcheb:
np it’s a nice way to poke around to see how it all works :)

hope you had a good new year

PR #104, opened 2022-08-17
Streampipes 577

dominikriemer:
### Remarks

Fixes: <Jira issue link>


streampipes-extensions/streampipes-connect-adapters-iiot/src/main/java/org/apache/streampipes/connect/iiot/protocol/stream/FileStreamProtocol.java
```diff
@@ -211,11 +211,12 @@ public GuessSchema getGuessSchema() throws ParseException {
 
     List<byte[]> dataByte = parser.parseNEvents(dataInputStream, 2);
 
-    EventSchema eventSchema = parser.getEventSchema(dataByte);
-
-    GuessSchema result = SchemaGuesser.guessSchma(eventSchema);
-
-    return result;
+    if (parser.supportsPreview()) {
```
tenthe, line 214: Are there parsers who do not support the preview? Might it be possible to remove this function in the future?
If so, could we mark it as Deprecated

tenthe (approved):

streampipes-extensions/streampipes-connect-adapters-iiot/src/main/java/org/apache/streampipes/connect/iiot/protocol/stream/FileStreamProtocol.java
dominikriemer: Yes, needed to distinguish between formats that support previews and those which do not. The interface has a default implementation pointing to false, so that no changes are required to existing formats. We can deprecate the method once all formats support the preview feature.

tenthe: I was wondering if it makes sense to deprecate it now, as a reminder to implement it for all parsers and remove it in the future releases?

PR #114, opened 2022-10-10
Use the Pulsar reader to fetch messages in Pulsar adapter

RobertIndie:
### Purpose
Currently, we are using the pulsar consumer to fetch messages in the pulsar adapter. This will create some redundant subscriptions in the pulsar broker. It's better to use the Pulsar reader to fetch small amounts of messages. 

### Approach
* Use Pulsar reader to fetch messages
* Use a more robust style to read messages instead of using `thread.sleep`


tenthe:
Thanks for providing a more robust implementation for the Pulsar adapter

PR #126, opened 2022-10-24
[STREAMPIPES-604] Add NGINX volume to compose deplyment

obermeier:
Sometimes it is usefull to change the configururation of the NGINX-Proxy.

E.g. to provide REST-Endpoints on a different port

### DIFF
--- a/installer/cli/deploy/standalone/ui/docker-compose.yml
+++ b/installer/cli/deploy/standalone/ui/docker-compose.yml
@@ -22,6 +22,8 @@ services:
     depends_on:
       - couchdb
       - consul
+    volumes:
+      - nginx:/etc/nginx/
     logging:
       driver: "json-file"
       options:
@@ -29,6 +31,9 @@ services:
         max-file: "1"
     networks:
       spnet:
+      
+volumes:
+  nginx:
 
 networks:
   spnet:


dominikriemer:
Cool! Should we also add the volume to the docker-compose files in installer/compose?

tenthe:
When users want to use SSL, they also have to change the nginx configurations in the docker-compose file, as described in [Documentation](https://streampipes.apache.org/docs/docs/deploy-use-ssl.html).
Do we have to consider or adjust anything here with the new volume mapping?

obermeier:
Is it a good solution to store the configuration and certificates in the same volume?

tenthe:
Good question, I am not sure. 
The location of the certificate is defined in /ui/nginx_config/ssl.conf.
I guess we can just change it there if we want to use two different volumes.
What do you think?

obermeier:
> Good question, I am not sure. The location of the certificate is defined in /ui/nginx_config/ssl.conf. I guess we can just change it there if we want to use two different volumes. What do you think?

Yes I like this idear

obermeier:
I realized that the volumes in k8s overrides the files from the docker image. (I should have known it.)
In this case the configuration must be placed to the volume via ConfigMap, SideCar, copy from ...
Since there is a good default configuration and adding a volume is easy this task could be left to the user!?
If he wants to change the configuration and persist this changes.

For this reason I removed the volume for the nginx configuration

PR #128, opened 2022-10-26
Improve the integration test framework

RobertIndie:
### Purpose

Improve the integration test framework

### Approach

* Refine the interface of `AdapterTesterBase`
* Use the low level approach to start and stop the adapter
* Add `DebugAdapterSInk` to collect the data for verification
* Improve the pulsar adapter integration tests

### DIFF
--- a/streampipes-connect/src/main/java/org/apache/streampipes/connect/adapter/Adapter.java
+++ b/streampipes-connect/src/main/java/org/apache/streampipes/connect/adapter/Adapter.java
@@ -18,6 +18,7 @@
 
 package org.apache.streampipes.connect.adapter;
 
+import com.google.common.annotations.VisibleForTesting;
 import org.apache.streampipes.connect.adapter.model.pipeline.AdapterPipeline;
 import org.apache.streampipes.connect.adapter.preprocessing.elements.SendToJmsAdapterSink;
 import org.apache.streampipes.connect.adapter.preprocessing.elements.SendToKafkaAdapterSink;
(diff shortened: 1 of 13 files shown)


tenthe:
Thanks for providing the PR. I really like the new version for the test framework, especially the `DebugAdapterSink` that can be used to get the resulting event via a `BlockingQueue`.

I tried to run the test locally but got a NullPointer exception in PulsarAdapterTester in line 111. The method takeEvent returns null.
I ran it without the env variable `TEST_MODE` starting a new docker service with Pulsar. Do you have any ideas what the problem might be?

Once I have a running local version I can also help you to refactor the part in the prepareAdapter method. Therefore, we can use the builder class `SpecificDataStreamAdapterBuilder`. We probably have to extend it a little bit to also directly add the values for the configurations.

RobertIndie:
Hi @tenthe . Thanks for your testing.

> I ran it without the env variable `TEST_MODE` starting a new docker service with Pulsar. Do you have any ideas what the problem might be?

I find that there is a race condition issue in the PulsarProtocol. When the `PulsarProtocol.run` returns, the pulsar consumer may not be set up completely. We need to wait for the subscribe operation to complete before `PulsarProtocol.run` returns. I think the best way for it is to do some refactor for the PulsarProtocol by using the CompletableFuture. I will do it in another PR. This PR only lets the consumer subscribe to the earliest position to avoid this problem. This is just a temporary solution because it changes the default configuration. Also, I will submit a new PR to make the subscription position configurable.

>  Therefore, we can use the builder class `SpecificDataStreamAdapterBuilder`.

Great! Thanks for your information and your help. Do you want to submit the patch to this PR or create a new PR?

PR #130, opened 2022-11-01
[STREAMPIPES-612] Add builder pattern for GenericAdapterStreamDescrip…

tenthe:
### Purpose
Until now, it was quite cumbersome to create a new AdapterDescription because we didn't provide a builder for that.

### Approach
Add a builder class `GenericDataStreamAdapterBuilder` to generate a new `GenericDataStreamAdapterDescription`

### Remarks

Related to [STREAMPIPES-612]

### DIFF
--- a/streampipes-connect/src/main/java/org/apache/streampipes/connect/adapter/format/json/object/JsonObjectFormat.java
+++ b/streampipes-connect/src/main/java/org/apache/streampipes/connect/adapter/format/json/object/JsonObjectFormat.java
@@ -18,14 +18,15 @@
 
 package org.apache.streampipes.connect.adapter.format.json.object;
 
-import org.apache.streampipes.connect.api.IFormat;
 import org.apache.streampipes.connect.adapter.format.json.AbstractJsonFormat;
+import org.apache.streampipes.connect.api.IFormat;
 import org.apache.streampipes.model.connect.grounding.FormatDescription;
 import org.apache.streampipes.sdk.builder.adapter.FormatDescriptionBuilder;
 
 public class JsonObjectFormat extends AbstractJsonFormat {
 
   public static final String ID = "https://streampipes.org/vocabulary/v1/format/json/object";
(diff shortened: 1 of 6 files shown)


tenthe:
@RobertIndie as discussed in #128 I added a builder to generate the adapter description.
However, I am not sure if this is already the best solution, because we have to 'recreate' the description that is already defined in the `declareModel` method of the adapter. 
Maybe we can find a way to only provide the user input programmatically for all the parameters.
What do you think?

RobertIndie (approved):

RobertIndie:
> Maybe we can find a way to only provide the user input programmatically for all the parameters.

Sounds like a good idea. But I wonder what the `declareModel` method of adapters is used for currently.

tenthe:
@RobertIndie yes, you are absolutely right. We have been quite inconsistent in the past. 
We have discussed this issue on the mailing list as well. 
I have enabled the checkstyle plugin for my IDE and tried to reformat the code before publishing it, but there might still be issues with my local configuration.
What is your opinion, should we include the code style changes gradually or should we automatically reformat the whole code base?
My idea was to gradually use a consistent code style to learn what is the best practice for the project. 
I'm happy to discuss this topic further, as I'm not sure how best to handle it with the existing code base.

RobertIndie:
@tenthe I see that we already have the checkstyle convention here: https://github.com/apache/incubator-streampipes/blob/dev/tools/maven/checkstyle.xml If we are all still +1 for this convention, we can enforce the checkstyle for each module gradually. We may need to reformat the code in serval PRs. And these PRs should only care about the reformatting.

BTW, I'm +1 for merging this PR. We can still reformat these codes in another PR.

tenthe:
@RobertIndie regarding the checkstyle, I do not have a strong opinion on which checkstyle configuration we use. I think it is important that we have a consistent code base. So I'm +1 for enforcing it gradually module by module.

PR #134, opened 2022-11-08
First draft python setup

kulbachcedric:
### Purpose
First draft python setup

### DIFF
--- a/streampipes-client-python/Makefile
+++ b/streampipes-client-python/Makefile
@@ -1,0 +1,13 @@
+COMMIT_HASH := $(shell eval git rev-parse HEAD)
+
+format:
+	pre-commit run --all-files
+
+doc:
+	mkdocs build
+
+livedoc: doc
+	mkdocs serve --dirtyreload
+
+rebase:
+	git fetch && git rebase origin/master
--- a/streampipes-client-python/.gitignore
+++ b/streampipes-client-python/.gitignore
@@ -1,0 +1,261 @@
+# Own directories
+.vscode
+.build
+
+# Created by https://www.toptal.com/developers/gitignore/api/visualstudiocode,pycharm,jupyternotebooks,python
+# Edit at https://www.toptal.com/developers/gitignore?templates=visualstudiocode,pycharm,jupyternotebooks,python
+
+### JupyterNotebooks ###
+# gitignore template for Jupyter Notebooks
+# website: http://jupyter.org/
+
+.ipynb_checkpoints
+*/.ipynb_checkpoints/*
+
+
+# IPython
+profile_default/
+ipython_config.py
+
+# Remove previous ipynb_checkpoints
+#   git rm -r .ipynb_checkpoints/
+
+### PyCharm ###
+# Covers JetBrains IDEs: IntelliJ, RubyMine, PhpStorm, AppCode, PyCharm, CLion, Android Studio, WebStorm and Rider
+# Reference: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839
+
+# User-specific stuff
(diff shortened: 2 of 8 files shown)


bossenti:
@kulbachcedric do we also want to use/provide `requiremts.txt` and `requirements_test/dev.txt` files? Or work with e.g. `poetry` as a dependency manager? what do you think?

streampipes-client-python/Makefile
```diff
@@ -0,0 +1,13 @@
...
+	pre-commit run --all-files
+
+doc:
+	mkdocs build
+
+livedoc: doc
+	mkdocs serve --dirtyreload
+
+rebase:
+	git fetch && git rebase origin/master
```
dominikriemer, line 13: @kulbachcedric when is this rebase called? Sorry, I'm totally unaware of the python stuff ;-)

kulbachcedric: > @kulbachcedric when is this rebase called? Sorry, I'm totally unaware of the python stuff ;-)

It's just for simplicity... We can remove it :-)

dominikriemer:
rat fails with three files without license header:
streampipes-client-python/README.md
  streampipes-client-python/.pre-commit-config.yaml
  streampipes-client-python/Makefile

Is it possible to add headers to these files as well?

kulbachcedric:
> rat fails with three files without license header: streampipes-client-python/README.md streampipes-client-python/.pre-commit-config.yaml streampipes-client-python/Makefile
> 
> Is it possible to add headers to these files as well?

Done ☺️

streampipes-client-python/Makefile
dominikriemer: We can keep it if it makes things easier...just wondering when this command would be running? 
@bossenti should this point to dev then?

bossenti: Ah yes, I our case it should point to `dev`.
It's just a convenient command that you can run locally. 
Nothing specific for python.

dominikriemer:
licensing looks good from my view! Thanks @kulbachcedric!

bossenti:
> > @kulbachcedric do we also want to use/provide `requiremts.txt` and `requirements_test/dev.txt` files? Or work with e.g. `poetry` as a dependency manager? what do you think?
> 
> I think using `requirements_*.txt` files is misleading as we want the package to be installable over pip. Unfortunately, I don't have any experience with poetry yet, but if I understand it correctly and if we write the tests for all python versions 3.8 upwards, poetry should be superfluous, right?

Yes, I fully agree with this part `as we want the package to be installable over pip` but in my point of my we still need a way to easily set up your local python environment. Do I miss anything here? From my current understanding one would need to install all requirements defined in the `setup.py` by hand or is there any common way I don't know so far?

RobertIndie:
> 
> Yes, I fully agree with this part `as we want the package to be installable over pip` but in my point of my we still need a way to easily set up your local python environment. Do I miss anything here? From my current understanding one would need to install all requirements defined in the `setup.py` by hand or is there any common way I don't know so far?

@bossenti We can define all required dependencies in the `setup.py` and then it will install them automatically when we run the `setup.py`. In this way, we don't need `requirements.txt` anymore. For more information, you can refer here: https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#declaring-required-dependency

bossenti:
Ah thanks @RobertIndie, I missed that you can also use the `setup.py` for installing the required packages locally without building the python package (see more at https://stackoverflow.com/questions/1471994/what-is-setup-py). Thanks for the hint :) 

So from my point of view, we are ready to merge :)

PR #142, opened 2022-11-17
Feature/streampipes 627

bossenti:
### Purpose

Integrate style and codecheks as well as unit tests to our CI (pr-validation).

### Remarks

Fixes: [STREAMPIPES-627](https://issues.apache.org/jira/browse/STREAMPIPES-627)

### DIFF
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -36,6 +36,7 @@ jobs:
 
   run-cypress-e2e-tests:
     runs-on: ubuntu-latest
+    needs: [python-quality-checks, python-unit-tests]
     steps:
       - name: Checkout
         uses: actions/checkout@v1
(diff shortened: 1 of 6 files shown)


tenthe:
Hi Tim,
the PR looks good.
One general question, do you already know how long the code validation checks will run?
If this is rather fast, we could run all code validation first, then run the unit tests and finally run the e2e tests.
What do you think?

dominikriemer (approved):

bossenti:
@tenthe `python-quality-check` took one minute and `unit-tests` between 30 und 60 seconds.
Currently, all jobs that are part of `pr-validation` run in parallel. So what would you change? I didn't get your suggestion, I'm afraid.

tenthe:
ok great. 
The `run-cypress-e2e-tests` takes about 30 minutes. 
My question basically was if we should wait for all the code validations to pass before triggering the e2e tests?

bossenti:
Indeed, thats a good one.
I'll adapt the workflow correspondingly.

bossenti:
@tenthe Jobs run now sequentially, see: https://github.com/apache/streampipes/actions/runs/3499092709

PR #147, opened 2022-11-19
chore/STREAMPIPES-590

bossenti:
### Purpose

Add licencse headers to all `strings.en` files.

### Remarks

Fixes: [STREAMPIPES-590](https://issues.apache.org/jira/browse/STREAMPIPES-590)

### DIFF
--- a/archetypes/streampipes-archetype-extensions-jvm/src/main/resources/archetype-resources/src/main/resources/__package__.pe.__packageName__.processor/strings.en
+++ b/archetypes/streampipes-archetype-extensions-jvm/src/main/resources/archetype-resources/src/main/resources/__package__.pe.__packageName__.processor/strings.en
@@ -1,20 +1,19 @@
-<!--
-  ~ Licensed to the Apache Software Foundation (ASF) under one or more
-  ~ contributor license agreements.  See the NOTICE file distributed with
-  ~ this work for additional information regarding copyright ownership.
-  ~ The ASF licenses this file to You under the Apache License, Version 2.0
-  ~ (the "License"); you may not use this file except in compliance with
-  ~ the License.  You may obtain a copy of the License at
-  ~
-  ~    http://www.apache.org/licenses/LICENSE-2.0
-  ~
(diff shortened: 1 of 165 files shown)


tenthe:
Hi Tim,
great job with adding all the license headers. 
Did this directly work or did you also have to change something in the java implementation when reading the files?

I think we can now also remove the <exclude> from the pom.xml file (line 1719). The language files are currently excluded from the checks of the rat plugin.

bossenti:
@tenthe thats a great suggestion!
The `strings.*` files are now covered by the `rat` plugin and it directly discovered two files that I missed ;) 

Fortunately, there was no need to adapt how we reed the files.
I've tested it with one or two examples.

tenthe:
Ok great.
Do you know why the e2e tests failed before? Is it related to the changes of the PR?

bossenti:
They should now go through, there was a silly mistake 😵‍💫 
https://github.com/apache/streampipes/pull/147/commits/356052276dc89e9b6ee4e84f9c16204ef45efbb1

But the rat plugin is still failing, I will investigate 
Maybe we should exclude the artifact strings instead of adding the header? 🤔

dominikriemer:
Rat fails due to some Maven error...there seems to be some mirror down, just wait a few hours

Connect to repo.maven.apache.org:443 [repo.maven.apache.org/151.101.40.215] failed: Connection timed out (Connection timed out) -> [Help 1]

dominikriemer (commented):
@bossenti Thank you very much for adding these comments.
I am not sure about the comment syntax: As these files are basically properties files, shouldn't this be a "#" instead of HTML syntax?

bossenti:
Hm, good point @dominikriemer 
I was also tinkering between these two options... I can also switch to `#` comment style
Is there any advice from the Java Library we are using for reading the resource files?

dominikriemer:
Yes, I think the Java convention proposes hash or explanation mark:

https://en.m.wikipedia.org/wiki/.properties

Thanks!

bossenti:
Header style is changed to `#`

PR #149, opened 2022-11-20
Streampipes 566

dominikriemer:
### Purpose

Fix bug where the data download dialog ignored query-specific settings when selecting "download current query".
In addition, this PR improves code structure by extracting all download-specific features from the DataLakeManagementV4 class into smaller classes and deletes obsolete and unused classes.
The module needs more refactoring though.

### Remarks

Fixes: <Jira issue link>


streampipes-data-explorer/src/main/java/org/apache/streampipes/dataexplorer/DataLakeManagementV4.java
```diff
@@ -50,39 +50,11 @@
...
-import java.time.temporal.ChronoField;
-import java.time.temporal.ChronoUnit;
-import java.time.temporal.TemporalAccessor;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
```
tenthe, line 53: The wildcard import should be removed

tenthe (commented):
I like the changes and I totally agree, we should further simplify this API. This is a good step into this direction.
Maybe we can also delete DataLakeMeasureResourceV3 & DataLakeResourceV3 since it is now deprecated for some time.

streampipes-data-explorer/src/main/java/org/apache/streampipes/dataexplorer/v4/query/StreamedQueryResultProvider.java
```diff
@@ -0,0 +1,113 @@
...
+        .findFirst();
+  }
+
+  /**
+   * Replaces the field 'time' of the data result with the actual timestamp field name of the measurement
+   * @param measurement contains the actual timestamp name value
+   * @param dataResult the query result of the database with 'time' as timestamp field name
+   */
+  private void changeTimestampHeader(DataLakeMeasure measurement, SpQueryResult dataResult) {
+    dataResult.getHeaders().set(0, measurement.getTimestampFieldName());
```
bossenti, line 111: Can we rely on `time` being always the first header?

bossenti (commented):
@dominikriemer great work! As far as I can see, this is a great simplification.
I have only roughly skimmed the changes, as I am not yet very deep into this part of StreamPipes
But the general impression is pretty good!
If you want a more detailed review, I'm happy to dig a little deeper.

streampipes-data-explorer/src/main/java/org/apache/streampipes/dataexplorer/v4/query/StreamedQueryResultProvider.java
dominikriemer: Thanks, fixed!

PR #150, opened 2022-11-21
[STREAMPIPES-591] Prevent tests from failing with Java 17+

dominikriemer:
### Purpose

This PR fixes failing powermock tests when compiling with Java 17+.
I will create a separate ticket to remove Powermock in the first release after 0.90.0

### Remarks

Fixes: <Jira issue link>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1844,6 +1844,17 @@ IoT data streams.
 					</execution>
 				</executions>
 			</plugin>
+			<plugin>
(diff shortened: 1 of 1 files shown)


bossenti:
@dominikriemer If not already done, you can add it to https://issues.apache.org/jira/browse/STREAMPIPES-592. Powermock is the barrier why we cannot directly switch so junit5 as well

PR #152, opened 2022-11-23
Create data stream endpoint for python client

SvenO3:
### Purpose

Create a data stream endpoint for the python client. This makes it possible to get the meta information of data streams in python and is the foundation to be able to receive the live data in the future.
### Remarks

Fixes: <Jira issue link>


bossenti:
@SvenO3, thanks a lot 🙏🏼 
I'll have a look on this probably tomorrow.

tenthe:
Hi Sven,
thanks a lot for providing a PR, it looks good.
I updated the API to read streams in [#155].
Once the PR is merged you can also update the route in the python code.

streampipes-client-python/streampipes_client/client/client.py
```diff
@@ -102,6 +102,7 @@ def __init__(
         # provide all available endpoints here
         # name of the endpoint needs to be consistent with the Java client
         self.dataLakeMeasureApi = DataLakeMeasureEndpoint(parent_client=self)
+        self.dataStreamApi = SpDataStreamEndpoint(parent_client=self)
```
bossenti, line 105: @SvenO3 would you mind just calling the class `DataStreamEndpoint`? (same applies for `SpDataStream`) 
I'm more in favor of class names without this prefix, since we would then need to add it to almost every class.
What do you think?

SvenO3: Yes, we can change this. I just used the same names as in the StreamPipes API and this prefix is only added to the SpDataStream and SpDataSet there

PR #807, opened 2022-11-27
770 issue templates

bossenti:


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,115 @@
...
+
+  - type: dropdown
+    attributes:
+      label: Affected StreamPipes components
+      description: >
+        Which StreamPipes components are affected by this bug?
+        Where did the bug occur?
+      multiple: true
+      options:
+        - Backend
```
dominikriemer, line 58: Somehow the mapping from these components to "core/extensions" went wrong during issue migration.
Should we use these labels anyway?
E.g., "Core", "Extensions (pipeline elements and adapters)"?

bossenti: I have already cleaned them up and corrected the "wrong" assignments.
So we can use the labels as planned.
Are you missing any other options here?
Or in general another input field?

tenthe (approved):
@bossenti awsome work. I really like the structure.
I think we will find out what we are missing when we use it.

RobertIndie (approved):
Awesome! LGTM

PR #814, opened 2022-11-29
Enable check style for `streampipes-integration-tests`

RobertIndie:
### Purpose

Discussion: #815 

Master issue: #820

Enable check style for `streampipes-integration-tests`.

This is the first try to enable checkstyle for the repo. I use the module `streampipes-integration-tests` as the example here.
If we are good with this PR, then we can apply it to all other modules gradually.

### Approach
* Add the `Indentation` convention to the `checkstyle.xml`. I saw that we use the different Indentation convention in the repo. Here  I use the [`google_checks.xml`](https://github.com/checkstyle/checkstyle/blob/master/src/main/resources/google_checks.xml) as the Indentation convention, which works for most of our codes.
* Change `checkstyle-header` to make it compatible with most of codes.
* Reformat the code in `streampipes-integration-tests`

### Validation

1. Make sure you are at the root path of the project
2. Run `maven validate -f streampipes-integration-tests`


tenthe (approved):

dominikriemer (approved):

bossenti (approved):
thanks a lot for taking over the initiative @RobertIndie 🙏🏼 
Do we want to create an issue to track the progress with validating all modules?

RobertIndie:
> Do we want to create an issue to track the progress with validating all modules?

Great idea. I have created an issue to track it: https://github.com/apache/streampipes/issues/820

PR #816, opened 2022-11-29
Let discussions also happen in the dev mailing list

RobertIndie:
### Purpose
Currently, the Github discussion doesn't seem to be synced to the dev mailing list. Let's make it also happen in the mailing list.


bossenti:
Hi Zike,

thats a great idea!

Issue #822, opened 2022-11-30
Ignore case when logging in with email address


dominikriemer:
### Body

Currently, correct username validation on the login page requires an exact match between the provided mail address and the registered address. 
Mail address validation should ignore case instead.

### StreamPipes Committer

- [X] I acknowledge that I am a maintainer/committer of the Apache StreamPipes project.

PR #823, opened 2022-11-30
Ignore case when fetching users from database (#822)

dominikriemer:
### Remarks

Fixes: <GitHub issue link>


tenthe:
Hi @dominikriemer,
do I understand this change correclty?

- The username in the database may contain capital letters
- The `UpdateUsernameViewMigration` script updates the **disign document** to always return the user name in lower case (and user names of old systems are not changed)
- Before a username (provided as a method parameter) is used, it is changed to lower case

dominikriemer:
Yes exactly ;-)

tenthe (approved):

PR #827, opened 2022-12-01
Enable checkstyle for streampipes-messaging-* modules

dominikriemer:
### Remarks

Fixes: <GitHub issue link>


dominikriemer:
@RobertIndie @tenthe I deactivated the ``NewlineAtEndOfFile`` rule as this caused problems on windows and proper line ending conversion between LF and CRLF should be ensured by Git. Is this ok?

RobertIndie (approved):

RobertIndie:
Hi @dominikriemer 
>  I deactivated the ``NewlineAtEndOfFile`` rule as this caused problems on windows and proper line ending conversion between LF and CRLF should be ensured by Git. Is this ok?

I am okay with that. It should be controlled by the git.

tenthe:
@dominikriemer I think that should be fine. 
The main intention of this rule is to make it easier for others and diff tools to add and track changes as it is described in [1].
I would say it is more important to have an approach that works on all development machines regardless of the operating system.

[1] https://checkstyle.sourceforge.io/apidocs/com/puppycrawl/tools/checkstyle/checks/NewlineAtEndOfFileCheck.html

tenthe (approved):

Issue #805, opened 2022-11-27
Create release notes with GitHub issues


bossenti:
Configure creation of release notes based on Github issues and labels.

PR #829, opened 2022-12-03
[#805] create custom release notes configuration

bossenti:
### Purpose

Create config file for custom release notes automatically created by GitHub

### Remarks

Fixes: #805

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,21 @@
+changelog:
+  exclude:
+    labels:
+      - dependencies
+  categories:
+    - title: New Features 🚀
+      labels:
+        - enhancement
+    - title: Bug fixes 🧰
+      labels:
+        - bug
+    - title: Breaking changes 💣
+      labels:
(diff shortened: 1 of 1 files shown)


bossenti:
See an example in my private streampipes repository: https://github.com/bossenti/incubator-streampipes/releases/tag/release-demo

bossenti:
One problem with this solution is that all issues/PRs related to the release tag are taken into account. This means that we would currently consider changes that relate to a future release but have already been merged. Can we live with that?

tenthe:
Is it possible to manually edit the release notes after they were automatically created?

bossenti:
Yes it is.
They are simly inserted into the release form you are about to create and you can modify them how ever you like.

tenthe:
Fine, but then we can remove it manually if we don't want a change to become part of the release. 
And since these features are not yet fully completed, we will see them in changes in the next release.

dominikriemer (approved):

PR #846, opened 2022-12-04
Add example files python client

bossenti:
### Purpose

Provide some tutorials for the Python client.

### Remarks


bossenti:
@dominikriemer @tenthe any idea how to properly put the Apache License header to Jupyter Notebooks?

tenthe:
I am not sure if this is even possible. It looks like it is json, right?
I saw one solution that suggested to add a dedicated json field for the license information, but I do not think that is a standard way. If it is not possible, we have to exclude it in the `apache-rat-plugin` configuration in the`pom.xml` file. There we already exluce `.csv` & `.json` files.

bossenti:
> I am not sure if this is even possible. It looks like it is json, right?

@tenthe yes, exactly `.ipynb` is basically JSON.
I will try to add the license as meta information and exclude jupyter notebooks in the `rat-plugin`

tenthe (approved):

PR #851, opened 2022-12-05
Simplify checkstyle configuration

RobertIndie:
### Remarks


tenthe (approved):

bossenti (commented):
just out of curiosity: what's the difference here? :)

RobertIndie:
> just out of curiosity: what's the difference here? :)

Because we can share the checkstyle configuration across the whole project by the plugin management in the parent pom file. Therefore, now we can just only add these lines to enable the checkstyle for every sub-modules:
```xml
<plugin>
  <groupId>org.apache.maven.plugins</groupId>
  <artifactId>maven-checkstyle-plugin</artifactId>
</plugin>
```

Issue #820, opened 2022-11-30
Enable check style for the project


RobertIndie:
### Body

This is the issue to track the progress of check-style enablement for all modules. 
If you want to take over some of the modules, please feel free to leave a comment here. And thanks for all your contribution!  :smiley:

Only the first level of sub-modules are listed here. If you feel that some of the modules(eg. `streampipes-extensions`) are heavy, please continue to split them into smaller tasks in this issue.

- [x] #830
- [x] #832
- [x] #833
- [x] #834
- [x] streampipes-service-discovery-api https://github.com/apache/streampipes/pull/845
- [x] streampipes-logging https://github.com/apache/streampipes/pull/847
- [x] streampipes-model-shared https://github.com/apache/streampipes/pull/847
- [x] streampipes-vocabulary https://github.com/apache/streampipes/pull/869
- [x] streampipes-model https://github.com/apache/streampipes/pull/847
- [x] streampipes-model-client https://github.com/apache/streampipes/pull/847
- [x] streampipes-test-utils https://github.com/apache/streampipes/pull/865
- [x] streampipes-serializers-json https://github.com/apache/streampipes/pull/865
- [x] streampipes-service-discovery-consul https://github.com/apache/streampipes/pull/845
- [x] streampipes-service-discovery https://github.com/apache/streampipes/pull/845
- [x] streampipes-service-base https://github.com/apache/streampipes/pull/845
- [x] streampipes-config https://github.com/apache/streampipes/pull/863
- [x] streampipes-messaging https://github.com/apache/streampipes/pull/827
- [x] streampipes-messaging-kafka https://github.com/apache/streampipes/pull/827
- [x] streampipes-storage-api https://github.com/apache/streampipes/pull/865
- [x] streampipes-storage-couchdb https://github.com/apache/streampipes/pull/865
- [x] streampipes-storage-management https://github.com/apache/streampipes/pull/866
- [x] streampipes-data-explorer https://github.com/apache/streampipes/pull/866
- [x] streampipes-connect-api https://github.com/apache/streampipes/pull/867
- [x] streampipes-dataformat https://github.com/apache/streampipes/pull/864
- [x] streampipes-sdk https://github.com/apache/streampipes/pull/867
- [x] streampipes-rest-shared https://github.com/apache/streampipes/pull/867
- [x] streampipes-container https://github.com/apache/streampipes/pull/867
- [x] streampipes-dataformat-cbor https://github.com/apache/streampipes/pull/864
[...]

PR #853, opened 2022-12-07
Fix the checkstyle plugin doesn't fail on violation

RobertIndie:
### Purpose
Fix https://github.com/apache/streampipes/issues/820#issuecomment-1339663188
The root cause is that the plugin configuration of the submodule is inherited from the parent project. We have skipped `failOnViolation` in the parent pom. Therefore the submodule also skips that.

This PR makes the checkstyle configuration of the parent module not be inherited to the submodule.
After we complete this issue: https://github.com/apache/streampipes/issues/820, we can turn it back and let the checkstyle configuration of the parent module apply to the whole project.

### Remarks

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1810,12 +1810,16 @@ IoT data streams.
 					<encoding>UTF-8</encoding>
 				</configuration>
 			</plugin>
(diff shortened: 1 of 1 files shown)


bossenti (approved):

tenthe:
Hi @RobertIndie, thanks a lot. I'll directly merge it, so I can use the changes for my PR

Issue #854, opened 2022-12-07
Implement StreamPipesFunctions for Python Client


SvenO3:
### Body

Create an implementation for the StreamPipesFunctions in python. 
The functions allow users to get the data of a StreamPipes data streams easily and makes it possible to work with the live data in python.
The implementation should be similar to the Java implementation and compatible with all messaging protocols.

### StreamPipes Committer

- [X] I acknowledge that I am a maintainer/committer of the Apache StreamPipes project.

PR #855, opened 2022-12-07
[#854] First implementation of the StreamPipesFunctions in python

SvenO3:
### Purpose

This is a first approach to implement the StreamPipesFunctions  and it allows to get the live data of data streams in python.
At the moment it only supports the Nats messaging protocol.

### Remarks


bossenti:
@SvenO3 congratulations to your first PR as a StreamPipes comitter 🥳 
Looks very promising, I'll have a deeper look on it the upcoming days!

streampipes-client-python/streampipes_client/functions/broker/broker.py
```diff
@@ -0,0 +1,96 @@
...
+
+from streampipes_client.model.resource.data_stream import DataStream
+
+
+class Broker(ABC):
+    """Abstract implementation of a broker.
+    A broker is used to subscribe to a data stream and to consume the published events.
+    """
+
+    async def connect(self, data_stream: DataStream, host_address: str) -> None:
```
bossenti, line 28: What is the intention behind making the `connect` method asynchronous? 
Usually, you should wait until the connection was established successfully, right?
Or do I miss anything here?

streampipes-client-python/streampipes_client/functions/broker/nats_broker.py
```diff
@@ -0,0 +1,81 @@
...
+
+        port: int
+            The port number of the connection.
+
+        Returns
+        -------
+        None
+        """
+        self.nats_client = await connect(f"nats://{host_address}:{port}")
+        if self.nats_client.connected_url is not None:
```
bossenti, line 51: else should raise an exception here

streampipes-client-python/streampipes_client/functions/function_handler.py
```diff
@@ -0,0 +1,192 @@
...
+    ----------
+    registration: Registration
+        The registration, that contains the StreamPipesFunctions.
+    client: StreamPipesClient
+        The client to interact with the API.
+    """
+
+    def __init__(self, registration: Registration, client: StreamPipesClient) -> None:
+        self.registration = registration
+        self.client = client
```
bossenti, line 57: I think this kind of breaks the basic idea of how we want to structure the API of the Python client.
But that's not really important for now, I'll give it some thought and come up with a solution.

streampipes-client-python/streampipes_client/functions/function_handler.py
```diff
@@ -0,0 +1,192 @@
...
+        Parameters
+        ----------
+        broker_name: str
+            A string that represents a broker.
+
+        Returns
+        -------
+        The broker which belongs to the name.
+        """
+        if broker_name == "nats":
```
bossenti, line 104: Would you mind to create an enum for supported brokers?

streampipes-client-python/streampipes_client/functions/broker/nats_broker.py
```diff
@@ -0,0 +1,81 @@
...
+from nats.aio.subscription import Subscription
+from streampipes_client.functions.broker.broker import Broker
+
+logger = logging.getLogger(__name__)
+
+
+class NatsBroker(Broker):
+    """Implementation of the NatsBroker"""
+
+    nats_client: Client
```
bossenti, line 31: This should be instance variables and not class variables right?

bossenti (commented):
@SvenO3 I left some comments on your PR.
I really appreciate your contribution, great work!
I look forward to refining and expanding it even more together

streampipes-client-python/streampipes_client/functions/broker/broker.py
SvenO3: Yes, the method waits until the connection is established. The `async` is just needed because you have to use the `await` keyword to connect to the broker. This is because the `connect` method in the nats library is a coroutine.

streampipes-client-python/streampipes_client/functions/broker/nats_broker.py
SvenO3: I don't really know if it's even possible that the else case happens because I think their should always be a `connected_url` if the connection was successful. I only added the if because the type checking failed as the url could be `None`.

streampipes-client-python/streampipes_client/functions/function_handler.py
SvenO3: I can do this. That's a good idea

streampipes-client-python/streampipes_client/functions/broker/nats_broker.py
SvenO3: That's right. Thanks for the detailed review

bossenti (approved):
Thanks a lot for the improvements @SvenO3 
That's a solid basement for our future work 👍🏼

PR #863, opened 2022-12-12
[#820] Enable checkstyle for streampipes-config and add suppress support for checkstyle

RobertIndie:
### Purpose
Master Issue: #820 

* Enable checkstyle for streampipes-config
* Add suppress support for checkstyle, now we can use this way to suppress the style check before the code:
```java
@SuppressWarnings("checkstyle:{Rule-name}")
...
```

I found that there is an issue in the IDEA: https://youtrack.jetbrains.com/issue/IDEA-80495/Settings-Operation-sign-on-next-line-does-not-work-for-automatically-breaking-strings.
Therefore, we seem couldn't reformat the code according to the `OperatorWrap` rule automatically. Haven't found a good solution yet. We could only format it manually.

### Remarks

### DIFF
--- a/streampipes-config/pom.xml
+++ b/streampipes-config/pom.xml
@@ -17,7 +17,8 @@
   ~
   -->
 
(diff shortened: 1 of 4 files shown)


bossenti (approved):
Really cool to have this decorator syntax 👍🏼

PR #875, opened 2022-12-20
[#771] add pre commit hook for linting and formatting

### LINKED ISSUE #771: UI linting as pre-commit hook
Create pre-commit support for all UI files

Imported from Jira [STREAMPIPES-631](https://issues.apache.org/jira/browse/STREAMPIPES-631). Original Jira may contain additional context.
Reported by: riemer.

smlabt:
Closes #771


ui/.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
```
bossenti, line 1: Please add an apache header here.

bossenti (changes requested):
thanks a lot for your contribution @smlabt 🙏🏼 
I'really like it.
Would you mind to create a very basic `developing.md` or something else where you quickly explain how to use and setup the pre-commit hook?

A general point for discussion: do we want to add the linting also as a check in our CI pipeline

ui/.husky/pre-commit
smlabt: Hi @bossenti,
thanks for your feedback :) I will add the missing apache header and also some documentation for the initial setup.

I'm also already working on a CI integration, but I think we should open a second PR for this. This second PR will be very large, because we have to "prettify" all existing files, in order to pass the CI pipeline. And I think this is not closely related to the pre commit hook. What's your opionion on this?

bossenti: great to here, +1 for having a dedicated PR for the reformatting
With regard to our Java codebase we apply the code style check module by module in #820.
Do you think splitting up would also make sense for the UI somehow although we don't have the same module structure there?

bossenti (approved):

RobertIndie (approved):

ui/.husky/pre-commit
tenthe: Hi, thanks a lot for the PR @smlabt.
I'm in favor of activating it module by module (if that is possible), since I'm not sure if a "full" automated reformatting might have some unexpected side effects.
I guess it makes sense to first get a bit more comfortable with the styling guidelines. What do you think?

tenthe (approved):

PR #881, opened 2022-12-22
Enable dependabot for ui and bump all minor versions

smlabt:
Is there a reason why package-lock.json is on .gitignore? Because I think otherwise we should check this in to our repository to have reproducible builds.

I am not familiar with dependabot, but according to the documentation (https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates#example-dependabotyml-file) this is how we can enable it for the UI.


tenthe:
I think you are right, the `package-lock.json` should not be in the .gitignore.
We already use dependabot to check the dependencies, but I did not know that it can further be configured via in `.github/dependabot.yml`. 
What do you think would be the best update strategy for the ui dependencies?

smlabt:
@tenthe I have checked the PRs opened by the dependabot (https://github.com/apache/streampipes/pulls?q=is%3Apr+is%3Aclosed+label%3Adependencies), but I do not see any UI related PRs. My hope is that with the `dependabot.yml` the bot will work.

If we want to reduce the amount of created PRs, we could try to ignore Patch Versions, as described here (https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#specifying-dependencies-and-versions-to-ignore) In this case the dependabot will open PRs for minor and major Updates only.

PR #892, opened 2022-12-25
fix: fix shebang with apache header

smlabt:
I think there is an issue with the pre-commit hook on windows, because the Shebang should be on the first line of the script. Because of the apache header it could not be interpreted correctly.

### DIFF
--- a/ui/.husky/pre-commit
+++ b/ui/.husky/pre-commit
@@ -1,3 +1,4 @@
+#!/usr/bin/env bash
 # Licensed to the Apache Software Foundation (ASF) under one or more
 # contributor license agreements.  See the NOTICE file distributed with
 # this work for additional information regarding copyright ownership.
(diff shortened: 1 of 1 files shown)


bossenti:
@aamirxshaikh could you please test if this works on your machine?

aamirxshaikh:
Sure @bossenti. I'd like to know what is the expected result that we are looking for.

Is it to check whether pre-commit hook works?

bossenti:
Exactly, just test whether the pre-commit hook can be executed. 
If I understood @smlabt correctly, this should now work on your end as well :)

bossenti:
To be more precise: try to commit without the `--no-verify` option

aamirxshaikh:
I have tested on my local, the commit works @bossenti

bossenti:
Great, thanks for testing!

PR #911, opened 2022-12-28
chore: remove peer dependencies and use postinstall

smlabt:
The two libraries "@streampipes/platform-services" and "@streampipes/shared-ui" are only available locally. So we can not use `npm install` without the flag `--legacy-peer-dependency`. Unfortunatelly dependabot uses `npm install` without this flag and therefore can not be used.

I think a good approach to manage these local libraries would be to use [Nx](https://nx.dev/) which brings support for angular monorepos (https://nx.dev/angular-tutorial/1-code-generation) But to use Nx we need to do some refactoring, I will open a discussion for this.

As a quick workaround I removed the two peer dependencies and added a "postinstall" Scrips which builds and installs the libraries after running `npm install`.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -29,6 +29,20 @@ updates:
     ignore:
       - dependency-name: "*"
         update-types: ["version-update:semver-patch"]
+  - package-ecosystem: "npm"
+    directory: "/ui/projects/streampipes/platform-services"
+    schedule:
+      interval: "daily"
+    ignore:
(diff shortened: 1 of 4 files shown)


dominikriemer (approved):

bossenti (approved):

bossenti:
great that you figured that out @smlabt 🚀 
I'm looking forward to the Nx discussion

tenthe (approved):

Issue #953, opened 2022-12-30
Improve handling of Consul env variables


dominikriemer:
### Body

Currently, the location of Consul to register services needs to be provided by a CONSUL_LOCATION env variable, which is defined in each Dockerfile for production usage and can be manually overridden. CONSUL_LOCATION implicitly requires the hostname and not the complete URL which might cause confusion.

* Introduce two new environment variables SP_CONSUL_HOST and SP_CONSUL_PORT
* Deprecate CONSUL_LOCATION
* Remove CONSUL_LOCATION env variable declaration from all Dockerfiles and use "consul" as default SP_CONSUL_HOST value 
* Use "localhost" as SP_CONSUL_HOST default if SP_DEBUG is set to true

### StreamPipes Committer

I acknowledge that I am a maintainer/committer of the Apache StreamPipes project.

PR #954, opened 2022-12-30
Improve handling of Consul env variables (#953)

dominikriemer:
### Remarks


dominikriemer:
@bossenti should we also replace CONSUL_LOCATION in the python client to use SP_CONSUL_HOST instead?

tenthe (approved):

bossenti:
> @bossenti should we also replace CONSUL_LOCATION in the python client to use SP_CONSUL_HOST instead?

If I am not mistaken, this is not yet used in the Python client. So we can directly use the new ENV name once we get to this point

PR #1014, opened 2023-01-02
Sp 1013

### LINKED ISSUE #1013: Update Java version to 17
As discussed in https://github.com/apache/streampipes/discussions/910 the goal is to update the java version from `11` to `17`.

<div type='discussions-op-text'>

<sup>Originally posted by **bossenti** December 28, 2022</sup>
Inspired by #740 I would like to discuss which Java versions we want to officially support.
These versions should be clearly communicated by us (e.g., in the README file) and also covered by our tests (see #841).
How do you like it if we focus only on the long-term support (LTS) versions of Java (currently: `8`, `11`, `17`)?
This is the way Apache Spark handles it, afaik.
Are there any other ideas/preferences?</div>

tenthe:
### Purpose

Update java version from version 11 to 17. Fixes #1013.

### Remarks


dominikriemer (approved):

bossenti (changes requested):
We should update the README as well

dominikriemer:
@bossenti thanks, I've updated the README - is there any other place you are aware of where we need to update the version?

bossenti (approved):

bossenti:
None that I'm aware of

Issue #1015, opened 2023-01-02
Bump Spring Boot to v3


dominikriemer:
### Body

- [x] Bump Spring Boot to 2.7, Spring to 5.3.24 and Spring Security to 5.8.1
- [x] Migrate Spring Security
- [x] Bump Spring Boot to 3.x
- [x] Use Java 17
- [x] Bump Spring and Spring Security to 6.x
- [x] replace javax modules by jakarta

### StreamPipes Committer

I acknowledge that I am a maintainer/committer of the Apache StreamPipes project.

PR #1016, opened 2023-01-02
Bump Spring versions, migrate Spring Security classes (#1015)

dominikriemer:
### Purpose

This PR bumps Spring Boot to 3.0.1 and Spring to 6.0.1, along with other dependency updates required for the migration.

### Remarks


tenthe (approved):
Thanks a lot Dominik. This is a big step

streampipes-extensions-management/src/main/java/org/apache/streampipes/extensions/management/connect/adapter/format/json/arraykey/JsonParser.java
```diff
@@ -30,11 +30,12 @@
 import org.apache.streampipes.model.schema.EventProperty;
 import org.apache.streampipes.model.schema.EventSchema;
 
-import jakarta.json.Json;
-import jakarta.json.stream.JsonParserFactory;
```
bossenti, line 34: With applying checkstyle to the whole code base this should not occur in that way, right?
Have we somewhere a misconfiguration?

bossenti (approved):
thanks a lot @dominikriemer for this huge work 🙏🏼 

Just to have it once asked: these `javax` to `jakarta` changes are compatible out of the box, right?

streampipes-extensions-management/src/main/java/org/apache/streampipes/extensions/management/connect/adapter/format/json/arraykey/JsonParser.java
dominikriemer: Thanks for reviewing!
Yes, the imports should be compatible.
I modified the checkstyle config so that there are no changes to the import order.

PR #1031, opened 2023-01-04
[hotfix] Remove unused message interface

tenthe:
### Purpose

I removed a package with the two interfaces `IMessageListener` & `IMessagePublisher` in the module `streampipes-commons`.
They are not used within our code base. 
@Reviewer, please confirm it is safe to delete.

### Remarks


bossenti (approved):
From my point of view it's fine to remove them
I can also do not find any usage of these interfaces.
Currently we have flagged it with `no release note`, do think this is valid @tenthe?
Or is it something that we want to communicate as `breaking change` in case this is used somewhere?

dominikriemer (approved):

tenthe:
I think this is some legacy code that is not used anymore. So I do not think we need it in the release notes

PR #1061, opened 2023-01-08
Sp 1058

dominikriemer:
### Purpose

This PR improves the streampipes model by removing obsolete or unused classes and removes two top-level objects `AbstractStreamPipesEntity` and `UnnamedStreamPipesEntity`. This reduces inheritance with the goal to make the model ready for cross-language data exchange.

### Remarks


streampipes-model/src/main/java/org/apache/streampipes/model/util/ElementIdGenerator.java
```diff
@@ -21,7 +21,7 @@
 
 public class ElementIdGenerator {
 
-  private static final String prefix = "urn:streampipes.apache.org:spi:";
```
tenthe, line 24: Could it be that we need an update script, or does this change only affect newly created IDs?

tenthe (approved):
The changes look good, I just had a question regarding the prefix in class `ElementIdGenerator` see comments.

streampipes-model/src/main/java/org/apache/streampipes/model/util/ElementIdGenerator.java
dominikriemer: Thanks for reviewing! This should not affect existing IDs.
I will create a migration script to cleanup existing models in the database as well.

bossenti (approved):

Issue #877, opened 2022-12-21
Apply UI linting to all modules


bossenti:
### Body

The following modules need to be included into the UI linting:

- [ ] cypress
- [ ] deployment
- [ ] projects
- [ ] src/app/_enums
- [ ] src/app/_guards
- [ ] src/app/_models
- [ ] src/app/add
- [ ] src/app/apidocs
- [ ] src/app/app-asset-monitoring
- [ ] src/app/app-overview
- [ ] src/app/assets
- [ ] src/app/configuration
- [ ] src/app/connect
- [ ] src/app/core
- [ ] src/app/core-model
- [ ] src/app/core-services
- [ ] src/app/core-ui
- [ ] src/app/CustomMaterial
- [ ] src/app/dashboard
- [ ] src/app/data-explorer
- [ ] src/app/editor
- [ ] src/app/files
- [ ] src/app/info
- [ ] src/app/login
- [ ] src/app/notifications
- [ ] src/app/NS
- [ ] src/app/pipeline-details
- [ ] src/app/pipelines
- [ ] src/app/profile
- [ ] src/app/services
- [ ] src/scss

Please update `ui/.eslintignore` and `ui/.prettierignore` if a module is updated.
Furthermore check to corresponding module in the list above and paste a link to the PR (see #820). 

### StreamPipes Committer

- [X] I acknowledge that I am a maintainer/committer of the Apache StreamPipes project.

aamirshaikh7:
Hey, I'm new to contributing to open source and I'm interested in contributing to this issue. How can I get started with it?

bossenti:
Hi @aamirshaikh7,
welcome to StreamPipes, great to hear that you want to contribute 🎉 

This issue depends on #876 and therefore it's preferable to have this PR merged before starting to work on this issue.
I expect this to happen in the upcoming days.

Nevertheless, I'll give a quick wrap up what to do:
1. Choose one of the UI models listed above and remove them from `ui/.eslintignore` and `ui/.prettierignore` locally
2. Run the UI commit hook like described in the [README](https://github.com/apache/streampipes/blob/dev/ui/README.md)
3. Fix the detected issues automatically (see README) or by hand
4. create a PR with your changes

In case you already want to contribute, feel free to have a look on our other issues.
There are some more [`good first issue`s](https://github.com/apache/streampipes/issues?q=is%3Aissue+is%3Aopen+label%3A%22good+first+issue%22).
If you are interested in UI development, you could for example have a look in #793

[...]

PR #1064, opened 2023-01-08
[#877] add formatting and linting to pipeline-details module

bossenti:
### Purpose

Apply formatting and linting to module `pipeline-details` 

### Remarks


ui/src/app/pipeline-details/components/edit/quickedit.component.html
```diff
@@ -16,52 +16,75 @@
...
   -->
 
-<sp-basic-nav-tabs [spNavigationItems]="tabs"
-                   [activeLink]="'quick-edit'"
-                   [showBackLink]="true"
-                   [backLinkTarget]="['pipelines']">
-
+<sp-basic-nav-tabs
+    [spNavigationItems]="tabs"
+    [activeLink]="'sp-quick-edit'"
```
bossenti, line 21: is it correct to adapt this reference as well?

tenthe: No, this should not be changed.
This is related to the route and not the component.
The route is registered in the module (see [1])

[1] https://github.com/apache/streampipes/blob/dev/ui/src/app/pipelines/pipelines.module.ts

tenthe (changes requested):

ui/src/app/pipeline-details/components/edit/quickedit.component.html
bossenti: Alright, then I'll revert it 🙂

tenthe (approved):

PR #1093, opened 2023-01-14
Sp 1085

### LINKED ISSUE #1085: Refactor `FileStreamProtocol`
### Body

The `FileStreamProtocol` can be used to replay event streams from files that are uploaded (e.g. csv file).
The protocol currently handles its own event schema and requires the `SendToBrokerReplayAdapterSink` class.
The logic of this class can be moved into the protocol so that it uses the regular connect API.

This issue should do some cleanup before we can start refactoring the connect API.

### StreamPipes Committer

tenthe:
### Purpose

Fixes #1085. 

Previously, the `FileStreamProtocol` required another sink (`SendToBrokerReplayAdapterSink`) for the adapter preprocessing pipeline. 
The reason for this was that the `FileStreamProtocol` needs to adjust the timestamp for correct replay, but it was the only adapter / protocol that used a different sink.
The `SendToBrokerReplayAdapterSink` class was deleted and the code was moved directly to the `FileStreamProtocol`.
Further, some minor refactoring was performed.

This should not be the final rewrite of the protocol, it is just a clean up step to better refactor the connect api.

### Remarks


streampipes-extensions-management/src/main/java/org/apache/streampipes/extensions/management/connect/adapter/model/pipeline/AdapterEventPreviewPipeline.java
```diff
@@ -79,4 +80,9 @@ public Map<String, GuessTypeInfo> makePreview() {
         .collect(Collectors.toMap(Map.Entry::getKey,
             e -> new GuessTypeInfo(e.getValue().getClass().getCanonicalName(), e.getValue())));
   }
+
+  @Override
+  public EventSchema getResultingEventSchema() {
+    return null;
```
dominikriemer, line 86: Why does this method always return null?

dominikriemer (approved):

streampipes-extensions-management/src/main/java/org/apache/streampipes/extensions/management/connect/adapter/model/pipeline/AdapterEventPreviewPipeline.java
tenthe: This is only temporary. In the future we probably will replace the class `AdapterEventPreviewPipeline` with a regular `AdapterPipeline` that uses a different sink.
I had to add it here to remove the dependency of the `FileStreamProtocol` with the internal adapter implementation.
If you have an idea on how to remove it, I'm happy to discuss it. If there is no quick solution, I would suggest to leave it as it is and remove it during the revision of the connect API.

Issue #1096, opened 2023-01-15
Improve structure of pipeline execution


dominikriemer:
### Body

Currently, the `PipelineExecutor` class executes several individual tasks when starting a pipeline. From the code structure, it is not clear which steps are executed when starting and stopping a pipeline.
To improve code structure, this class can be decomposed into individual tasks for each required step when executing a pipeline operation.

### StreamPipes Committer

I acknowledge that I am a maintainer/committer of the Apache StreamPipes project.

PR #1097, opened 2023-01-15
Improve structure of pipeline execution management (#1096)

dominikriemer:
### Remarks


streampipes-model/src/main/java/org/apache/streampipes/model/api/EndpointSelectable.java
```diff
@@ -0,0 +1,33 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.apache.streampipes.model.api;
+
+public interface EndpointSelectable {
```
tenthe, line 21: Can we add some descriptions to the interface methdos. I did not directly understand what `getDetachPath` was used for

tenthe (commented):
I had a minor comment regarding the description of an API.
Does the `CurrentPipelineElementProvider` return the running pipeline elements?

streampipes-model/src/main/java/org/apache/streampipes/model/api/EndpointSelectable.java
dominikriemer: I've added some comments!

`getDetachPath` provides the path of a processor, set or sink in the exensions module that must be called to detach the element.

tenthe (approved):

PR #1117, opened 2023-01-17
[#1114] Adapters can now be added without starting them

### LINKED ISSUE #1114: Create an adapter without starting it
## User Story
As a user, I want to configure an adapter, so that it is created but not started.

## Acceptance Criteria
- Adapter is added to `My adapters` view but not started
- E2E test validates this behaviour

## Additional Information
- When creating the adapter in the last view **Adapter Settings** there should be an option to only create the adapter without instantiating it.

### StreamPipes Committer

tenthe:
### Purpose

Fix #1114. Adapters can now be added without having to be started.
This PR also includes a fix for the FileStreamAdapter. When the adapter is stopped, the running replays are now interrupted.

### Remarks

### DIFF
--- a/ui/src/app/connect/components/adapter-configuration/start-adapter-configuration/adapter-options-panel/adapter-options-panel.component.ts
+++ b/ui/src/app/connect/components/adapter-configuration/start-adapter-configuration/adapter-options-panel/adapter-options-panel.component.ts
@@ -16,7 +16,7 @@
  *
  */
 
-import { Component, EventEmitter, Input, OnInit, Output } from '@angular/core';
+import { Component, EventEmitter, Input, Output } from '@angular/core';
 
 @Component({
     selector: 'sp-adapter-options-panel',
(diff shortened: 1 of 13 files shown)


bossenti (approved):

ui/src/app/connect/components/adapter-configuration/start-adapter-configuration/adapter-options-panel/adapter-options-panel.component.ts
```diff
@@ -36,6 +36,9 @@ export class SpAdapterOptionsPanelComponent {
     @Input()
     dataCy: string;
 
+    @Input()
+    isChecked = false;
```
dominikriemer, line 40: What is the default behaviour? Adapters are started upon creation?

dominikriemer: Ok, I see that `startAdapterNow` is true. Why is `isChecked` initially set to false?

tenthe: The default value is `false` only for the `AdapterOptionsPanel` component. This component is also used for other configurations like **Remove Duplicate**. Since the default value for all other configurations is `false`, I decided to set the default value of this component to `false` as well.

PR #1122, opened 2023-01-18
[#1121] Add replay once option to FileStreamProtocol

### LINKED ISSUE #1121: Add option for single file replay in FileStreamAdapter
## User Story
As a user, I want to have the options to replay a file only once and ignore the timestamp for the replay in the FileStreamAdapter.

## Acceptance Criteria
No specific e2e test is required as this functionality will replace the FileSet adapter in the tests.

## Additional Information
- An additional parameter is added to the FileStreamAdapter configuration indicating whether the file should be replayed only once or multiple times
- The default behaviour should be to continuesly replay the file

### StreamPipes Committer

tenthe:
### Purpose

- Fixes #1121
- Implements the ability for the `FileStreamAdapter` to replay a file only once
- Additional parameter added to the `FileStreamAdapter` configuration to enable this feature
- Default behavior set to continuously replay the file

### Remarks
@dominikriemer do we need an update script for old adapters using the `FileStreamAdapters`?
Since we will update the connection API, should we implement this update now or later for the new data model?

### DIFF
--- a/streampipes-extensions/streampipes-connect-adapters-iiot/src/main/java/org/apache/streampipes/connect/iiot/protocol/stream/FileStreamProtocol.java
+++ b/streampipes-extensions/streampipes-connect-adapters-iiot/src/main/java/org/apache/streampipes/connect/iiot/protocol/stream/FileStreamProtocol.java
@@ -54,6 +54,12 @@ import java.util.concurrent.TimeUnit;
 
 public class FileStreamProtocol extends Protocol {
 
+  private static final String REPLACE_TIMESTAMP = "replaceTimestamp";
+  private static final String SPEED = "speed";
+  private static final String FILE_PATH = "filePath";
+  private static final String REPLAY_ONCE = "replayOnce";
+
+
   private static final Logger logger = LoggerFactory.getLogger(FileStreamProtocol.class);
 
   public static final String ID = "org.apache.streampipes.connect.iiot.protocol.stream.file";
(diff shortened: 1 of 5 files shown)


streampipes-extensions/streampipes-connect-adapters-iiot/src/main/java/org/apache/streampipes/connect/iiot/protocol/stream/FileStreamProtocol.java
```diff
@@ -199,10 +274,16 @@ public ProtocolDescription declareModel() {
...
+        .requiredFile(Labels.withId(FILE_PATH), Filetypes.CSV, Filetypes.JSON, Filetypes.XML)
+        .requiredMultiValueSelection(Labels.withId(REPLACE_TIMESTAMP),
             Options.from(""))
-        .requiredFloatParameter(Labels.withId("speed"))
+        .requiredSingleValueSelection(Labels.withId(REPLAY_ONCE), Options.from("no", "yes"))
+        .requiredAlternatives(Labels.withId(SPEED),
+            Alternatives.from(Labels.withId(KEEP_ORIGINAL_TIME), true),
+            Alternatives.from(Labels.withId(FASTEST)),
+            Alternatives.from(Labels.withId(SPEED_UP_FACTOR),
+                StaticProperties.group(Labels.withId("speed-up-factor-group"),
```
bossenti, line 285: Do we want to use a variable here as well?

tenthe: Good point. I left it as a String because it is only used once, but I can change it to a variable

bossenti: I see, would just be for the sake of consistency

PR #1126, opened 2023-01-19
refactor: introduce messaging endpoint for python client

bossenti:
### Purpose

Introduce a new kind of kind endpoint.
The messaging endpoint is designed to community with the StreamPipes messaging layer.
One of the first implementations of such a messaging endpoint will be the function handler.
This PR prepares the refactoring of the function handler to better integrate with the existing code structure of the Python client.

### Remarks


streampipes-client-python/streampipes_client/endpoint/endpoint.py
```diff
@@ -209,3 +212,76 @@ def get(self, identifier: str) -> Resource:
...
+        )
+
+    @_broker.setter
+    def _broker(self, broker: Broker) -> None:
+        """Setter method for internal property `broker`"""
+        self.__broker = broker
+
+    @property
+    @abstractmethod
+    def _container_cls(self) -> Type[ResourceContainer]:
```
SvenO3, line 261: Is there a reason why you implemented the methode `_container_cls` again, even though the `MessagingEndpoint` also inherits it from the `Endpoint` class?

bossenti: Good catch, that's just silly 😅

tenthe (approved):

PR #1135, opened 2023-01-20
Sp 1133

dominikriemer:
### Purpose

Adds support to functions to create output data streams so that results of functions can be played back into StreamPipes.

### Remarks

The StreamPipes function API was slightly changed to return a `FunctionConfig` instead of a pure `FunctionId`.


bossenti (commented):
Would it make sense to include some tests here?

tenthe (approved):

Issue #1104, opened 2023-01-16
Add GitHub wofklow that tags PR


bossenti:
### Body

Implement a GitHub workflow that automatically assigns labels to PRs.
Fur this purpose consider to use the following GitHub action: https://github.com/actions/labeler

### ToDos:

- [x] provide config file to define labeling rules
- [x] implement GitHub workflow that runs for every created PR and assigns labels to it

### Metoring
As this ticket is marked as good first issue: one of @dominikriemer, @tenthe, or @bossenti are happy to provide help for getting started, just tag (one of) them if you want to start working on this issue and need some help.
It would be great if you discover our repository by yourself (this is a great oppurtinity to get to know the basic StreamPipes components) and come up with a proposal for labeling rules. But if you struggle with that we are happy to help you making a first draft.

### StreamPipes Committer

I acknowledge that I am a maintainer/committer of the Apache StreamPipes project.

Harry262530:
Hello,
I would like to contribute in this issue

bossenti:
Hi @Harry262530, awesome 🤩 
Do you need any help for getting started?

bossenti:
Okay, the first step is to fork this repository to your own GitHub account.
Next, check out your own fork and create a branch which should ideally be named `1104-add-github-wofklow-that-tags-pr` (or similar).
Then you already can start with implementing the GitHub workflow.
You can find all our existing workflows in [`.github`](https://github.com/apache/streampipes/tree/dev/.github).

Once you have the workflow ready, you can create a PR in this repository and we will review you it 
Feel free to reach out in case you get stuck :)

Harry262530:
sure
i'll try my best

Harry262530:
I need a little Guidance,
 I have created a branch of  this repo and I have thoroughly gone through labeler repo and workflow of this repo, now I am stuck at how to start working with the workflow and including labels.

bossenti:
Okay, so at first we want to have an additional workflow. So there is no need to integrate with the existing ones.
You can create a new file and call it e.g. `pr-labeling.yaml`.
This workflow should be executed when a PR is created.
Here you can follow the example of `labeler` pretty close: https://github.com/actions/labeler#create-workflow
The more difficult part is the configuration file. If you want to, you can take some time and familiarize your self with our existing labels and the codebase and come up with a proposal.
But you don't need to, you can also just provide the workflow itself and we will then come up with the configuration file.

Does this help you?

bossenti:
[...]

PR #1142, opened 2023-01-21
[#1104] add GitHub workflow to label PRs automatically

Harry262530:
### Purpose

Implement a GitHub workflow that automatically assigns labels to PRs.
Fur this purpose consider to use the following GitHub action: https://github.com/actions/labeler

### ToDos:

- [x] provide config file to define labeling rules
- [x] implement GitHub workflow that runs for every created PR and assigns labels to it

### Remarks


.github/prlabeller.yml
```diff
@@ -0,0 +1,21 @@
+# Add repo label to any change in any file
```
bossenti, line 1: This is just the example config from the GitHub action, right?
That does not suit for us, but that's not a problem.
I will take care of adapting the config accordingly.

.github/workflows/prlabelling.yaml
```diff
@@ -0,0 +1,16 @@
+name: "Pull Request Labeler"
+run-name: ${{ github.actor }} created Pull Request Labeler
+
+on:
+- pull_request_target
```
bossenti, line 5: If I am not mistaken, this event is only produced when a PR from a forked repository is created. 
This is not what we desire here. I think the more general `pull_request` event is better suited for us and we should filter on opening events. Read more about that here: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

config_merge.yml
```diff
@@ -0,0 +1,18 @@
+name: 'PRLabeler'
```
bossenti, line 1: What's the purpose of this file?
From my understanding, we only need the other two files.

.github/workflows/prlabelling.yaml
```diff
@@ -0,0 +1,16 @@
...
+
+jobs:
```
bossenti, line 7: It would be great if you could filter for PRs created by @dependabot
To do so you can follow this example [here](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#accessing-and-using-event-properties)

bossenti (commented):
Thanks lot for initiative @Harry262530 🙏🏼 
Don't worry about the configuration file, I can take care of that part, but I'm not sure when I'll find the time, so it might take a while.

.github/prlabeller.yml
Harry262530: yah,
basically i was trying to give at least one label to any issue as 'repo'

.github/workflows/prlabelling.yaml
Harry262530: thanks for the advice
should I change it and then send another pr?

config_merge.yml
Harry262530: actually for the past 1 day i studied about pull request and labels so I saw this file on one repo, so I included it

.github/workflows/prlabelling.yaml
bossenti: You can just leave this PR open and push new commits to your existing branch, then this PR is updated automatically

Harry262530: sure,
i'll do that

.github/prlabeller.yml
bossenti: I see, but we don't use such a label
You can find an overview of all our labels [here](https://github.com/apache/streampipes/labels)
Feel free to improve the config, if you want to e

config_merge.yml
bossenti: Can you provide me with an example?

.github/prlabeller.yml
Harry262530: sure,
i will try my level best

.github/workflows/prlabelling.yaml
Harry262530: should i assign permissions ,contents and pull request or just leave it for default

bossenti: I think this looks fine for now

.github/pr-labeler-config.yml
```diff
@@ -0,0 +1,46 @@
+backend:
+# TODO
```
bossenti, line 2: @tenthe @dominikriemer what do you think we should provide here?

bossenti:
@tenthe @dominikriemer I came with an initial version of the label config
Although, I'm not really user whether all these file path filters will work, I would just give it a try
What do you think?

.github/pr-labeler-config.yml
dominikriemer: Not sure if it's possible but we could filter for all modules starting with streampipes-* except streampipes-extensions and non-java modules

.github/pr-labeler-config.yml
```diff
@@ -0,0 +1,61 @@
...
+
+java:
+  - ./**/*.java
+
+javascript:
+  - ./**/*.js
+
+'pipeline elements':
+  - streampipes-exensions/streampipes-extensions-*/**/*
+  - streampipes-extensions/streampipes-pipeline-*/**/*
```
dominikriemer, line 47: Do we have such a module in streampipes-extensions currently?

dominikriemer (approved):
Looks good to me!

tenthe (approved):

PR #1144, opened 2023-01-22
[#877] Format and lint CustomMaterial module

aamirxshaikh:
### Purpose

Applying UI linting and formatting to the `CustomMaterial` module. Removing the `CustomMaterial` module from `ui/.eslintignore` and `ui/.prettierignore` and fixing formatting.

### DIFF
--- a/ui/.eslintignore
+++ b/ui/.eslintignore
@@ -21,7 +21,6 @@ cypress/fixtures/**/*.csv
 
 # Remove these in the future to lint additional modules
 # Please also see .prettierignore
-src/app/CustomMaterial
 src/app/dashboard
 src/app/data-explorer
 src/app/editor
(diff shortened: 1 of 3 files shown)


tenthe (approved):

bossenti:
Do we want to rename the module to `custom-material` @tenthe @dominikriemer? Would be consistent with the other module names

tenthe:
Yes, I think that is a good idea. 
We should also discuss if we still need this module. We should address the question whether it is good practice to make all dependencies into one module which is then imported by all others or if it is better to import the material modules directly.

bossenti:
👍🏼 
But that's a follow-up topic right? 🙂

tenthe:
Yes, that is a follow up topic

PR #1147, opened 2023-01-23
refactor: further clean up python data model & introduce function definition

bossenti:
### Purpose

This PR contains some minor refactorings:
* improve import paths of `broker`-classes
* restructure python model and reduce usage of `BaseElement` class for a more concise class hirarchy

Furthermore, it introduces the data model of a `FunctionDefinition` which is required for the further extension of the function handler.

### Remarks


SvenO3:
I also work on the python functions at the moment and try to support output data streams like in Java (see #1135). So I am changing the data model and definition of a function too, but I try to fit my changes with yours.

streampipes-client-python/streampipes_client/model/common.py
```diff
@@ -55,30 +55,33 @@ class BaseElement(BasicModel):
...
 
-class EventPropertyQualityRequirement(BaseElement):
+class EventPropertyQualityRequirement(BasicModel):
     """
     Data model of an `EventPropertyQualityRequirement` in compliance to the StreamPipes Backend.
     """
 
+    element_id: Optional[StrictStr]
     minimum_property_quality: Optional[BaseElement] = Field(alias="eventPropertyQualityDefinition")
     maximum_property_quality: Optional[BaseElement] = Field(alias="eventPropertyQualityDefinition")
```
SvenO3, line 65: The EventPropertyQualities and EventPropertyQualityRequirement aren't used by the API anymore and I think we could remove them. The elementId is also not used for every class anymore and we probably should change this. But I can do this in another PR too.

SvenO3 (approved):

streampipes-client-python/streampipes_client/model/common.py
bossenti: That would be great 👍

bossenti:
@SvenO3 cool 🤩 
Then let's align on this somewhen in the near future

PR #1149, opened 2023-01-23
Extend REST API by endpoint to get a specific function definition

bossenti:
### Purpose

Extend the REST API by the endpoint `/v2/functions/<functionId>` that returns the definition of the given function.

### Remarks


tenthe (approved):
The PR looks good. 
I will take a look at the e2e tests as they seem to fail on all PRs.

bossenti:
What would happen if I query by an unknown id? Would this result in an internal exception or does the framework catch the exception and return a proper API response?

tenthe:
That is a good point. How could we test it properly?
I guess the best way would be to mock `FunctionRegistrationService.INSTANCE` but since it is an enum I am not sure if it is easily possible. One option would be to change the implementation and inject an instance of `FunctionRegistrationService` into the `FunctionsResource`, then it would be easier to mock and test the REST interface.

PR #1165, opened 2023-01-26
More specific types for large numbers.

### LINKED ISSUE #1164: More specific types for numerical values when guessing schema

Connect

### What happened?

When the schema of data is guessed, many numerical formats are subsumed by the float type.
Integer, Double, Float, Long. [Code location](https://github.com/apache/streampipes/blob/a23795b43a4a3d67021cc961ea2d956b639e390f/streampipes-extensions-management/src/main/java/org/apache/streampipes/extensions/management/connect/adapter/format/util/JsonEventProperty.java#L58)

This leads sometimes to error  when the data does not fit into the range of float.

E.g. When a timestamp attribute is given in ms
`{
  "tree_size": 162451767, 
  "timestamp": 1674637204330, 
  "sha256_root_hash": "mIYlnASHw5nD\/9etlz\/BUoDhrAkER5H09IIK1I+jfmE="
 }`

In this case it is not possible to use this attribute as timestamp since an exception occurs when the data is converted.

What do you think to use more specific types?

### How to reproduce?

Guess a schema from data with contains a number greater than float.
E.g. https://ct.googleapis.com/logs/argon2023/ct/v1/get-sth

### Expected behavior

In this example the guessed type schould be of type Long

_No response_

obermeier:
fix #1164

### DIFF
--- a/streampipes-extensions-management/src/main/java/org/apache/streampipes/extensions/management/connect/adapter/format/util/JsonEventProperty.java
+++ b/streampipes-extensions-management/src/main/java/org/apache/streampipes/extensions/management/connect/adapter/format/util/JsonEventProperty.java
@@ -56,10 +56,14 @@ public class JsonEventProperty {
       resultProperty.setRuntimeName(key);
       ((EventPropertyPrimitive) resultProperty).setRuntimeType(XSD.STRING.toString());
     } else if (o.getClass().equals(Integer.class) || o.getClass().equals(Double.class)
-        || o.getClass().equals(Float.class) || o.getClass().equals(Long.class)) {
+        || o.getClass().equals(Float.class)) {
       resultProperty = new EventPropertyPrimitive();
       resultProperty.setRuntimeName(key);
       ((EventPropertyPrimitive) resultProperty).setRuntimeType(XSD.FLOAT.toString());
+    } else if (o.getClass().equals(Long.class)) {
+        resultProperty = new EventPropertyPrimitive();
+        resultProperty.setRuntimeName(key);
+        ((EventPropertyPrimitive) resultProperty).setRuntimeType(XSD.LONG.toString());
     } else if (o.getClass().equals(LinkedHashMap.class)) {
       resultProperty = new EventPropertyNested();
       resultProperty.setRuntimeName(key);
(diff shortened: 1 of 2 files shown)


bossenti (commented):
thanks a lot @obermeier 🙏🏼 
Your changes look good to me
Can you please try to address the code style issues: https://github.com/apache/streampipes/actions/runs/4016747501/jobs/6900267599#step:4:70961
You can read on how to apply our checkstyle config here: https://cwiki.apache.org/confluence/display/STREAMPIPES/Code+Style+-+Java

tenthe:
Hello @obermeier,
thanks for opening the issue and providing a PR.

I recently encountered the same issue and changed it in [PR-1099](https://github.com/apache/streampipes/pull/1099/files). However, I suspect that the changes only affect the CSV format, as I tried your example and got the same behavior. 

The default should be that we always assume a floating point number. If the number is too large for a float (e.g. a Unix timestamp in ms) it is assumed to be a double (see class 'DatatypeUtils'). We decided to do this because if the first value of the CSV file is represented as an integer 1 and not as 1.0, when another value is then a float (e.g. 2.4) we will get runtime exceptions when storing the data in the time series storage. 

Your implementation now implements the same behavior, correct?
Do your changes also affect the values at runtime or only the description of the event?

obermeier:
> thanks a lot @obermeier 🙏🏼 Your changes look good to me Can you please try to address the code style issues: https://github.com/apache/streampipes/actions/runs/4016747501/jobs/6900267599#step:4:70961 You can read on how to apply our checkstyle config here: https://cwiki.apache.org/confluence/display/STREAMPIPES/Code+Style+-+Java

Thank you for this hint and the nice description!

obermeier:
> Hello @obermeier, thanks for opening the issue and providing a PR.
> 
> I recently encountered the same issue and changed it in [PR-1099](https://github.com/apache/streampipes/pull/1099/files). However, I suspect that the changes only affect the CSV format, as I tried your example and got the same behavior.
> 
> The default should be that we always assume a floating point number. If the number is too large for a float (e.g. a Unix timestamp in ms) it is assumed to be a double (see class 'DatatypeUtils'). We decided to do this because if the first value of the CSV file is represented as an integer 1 and not as 1.0, when another value is then a float (e.g. 2.4) we will get runtime exceptions when storing the data in the time series storage.
> 
> Your implementation now implements the same behavior, correct? Do your changes also affect the values at runtime or only the description of the event?



> Hello @obermeier, thanks for opening the issue and providing a PR.
> 
> I recently encountered the same issue and changed it in [PR-1099](https://github.com/apache/streampipes/pull/1099/files). However, I suspect that the changes only affect the CSV format, as I tried your example and got the same behavior.
> 
> The default should be that we always assume a floating point number. If the number is too large for a float (e.g. a Unix timestamp in ms) it is assumed to be a double (see class 'DatatypeUtils'). We decided to do this because if the first value of the CSV file is represented as an integer 1 and not as 1.0, when another value is then a float (e.g. 2.4) we will get runtime exceptions when storing the data in the time series storage.
> 
> Your implementation now implements the same behavior, correct? Do your changes also affect the values at runtime or only the description of the event?

Hi @tenthe, I like the Idear to use just Floats and Doubles. A timestamp in ms shoud fit to this model.
I ralized that I could not reproduce my "bug".

Issue #1146, opened 2023-01-23
Session based implementation of IoTDB sink (fix issues in current JDBC based implementation)


SteveYurongSu:
### Apache StreamPipes version

dev (current development state)

### Affected StreamPipes components

Connect

### What happened?

Recently, I was trying to play around streampipes and found that iotdb sink could not work properly.

I checked the source code and found that the current JDBC based implementation of iotdb sink is outdated, especially for the current production versions of iotdb (0.13.x and 1.0.x).

I‘d like to contribute a brand new iotdb sink implementation, which has the following features:

- Support production ready versions of IoTDB (version >= 0.13.0)
- Support by iotdb-session API, which can provide higher throughput and lower latency 
- Support schema-less write: no need to create storage group (database) before insertion

### How to reproduce?

Graph:
Water Level 1 -> IoTDB

I debugged the iotdb sink in pipeline-element mode and found that the current iotdb sink cannot connect to the iotdb instance (version >= 0.13.x).

### Expected behavior

IoTDB sink should be used with IoTDB (version >= 0.13.0)

### Additional technical information

OS
```
Mac (Apple M1)
MacOS 13.0.1 (22A400)
```

Java
```
java version "17.0.6" 2023-01-17 LTS
Java(TM) SE Runtime Environment (build 17.0.6+9-LTS-190)
Java HotSpot(TM) 64-Bit Server VM (build 17.0.6+9-LTS-190, mixed mode, sharing)
```

Maven
```
Java version: 17.0.6, vendor: Oracle Corporation, runtime: /Library/Java/JavaVirtualMachines/jdk-17.jdk/Contents/Home
Default locale: zh_CN_#Hans, platform encoding: UTF-8
OS name: "mac os x", version: "13.0.1", arch: "aarch64", family: "mac"
```

### Are you willing to submit a PR?

Yes, I am willing to submit a PR!

dominikriemer:
Hi @SteveYurongSu, awesome, thanks for the great description!
Please ask if you have any questions.
We can also refactor the sink to a one-class structure by extending `StreamPipesDataSink` instead of the currently used model for the IoTDB sink.

PR #1166, opened 2023-01-26
[#1146] Session based implementation of IoTDB sink (fix issues in current JDBC based implementation)

SteveYurongSu:
### Purpose

Recently, I was trying to play around streampipes and found that iotdb sink could not work properly.

I checked the source code and found that the current JDBC based implementation of iotdb sink is outdated, especially for the current production versions of iotdb (0.13.x and 1.0.x).

I‘d like to contribute a brand new iotdb sink implementation, which has the following features:

- Support production ready versions of IoTDB (version >= 0.13.0)
- Support by iotdb-session API, which can provide higher throughput and lower latency
- Support schema-less write: no need to create storage group (database) before insertion

See #1146 for more detail :D

### Remarks

Old versions (< v0.13) of IoTDB are no longer supported.

### Tests
I setup a pipeline as the following:
image

Then I checked the data that IoTDB received:
image
image

### DIFF
--- a/installer/cli/deploy/standalone/iotdb/docker-compose.dev.yml
+++ b/installer/cli/deploy/standalone/iotdb/docker-compose.dev.yml
@@ -16,5 +16,8 @@
 version: "3.4"
 services:
   iotdb:
+    # 0.13.3 is the only recommended production ready version for edge side (standalone) deployment
+    image: apache/iotdb:0.13.3-node
     ports:
-      - 6667:6667
+      # rpc port
+      - "6667:6667"
--- a/installer/cli/deploy/standalone/iotdb/docker-compose.yml
+++ b/installer/cli/deploy/standalone/iotdb/docker-compose.yml
@@ -16,9 +16,10 @@
 version: "3.4"
 services:
   iotdb:
-    image: apache/iotdb
+    # 0.13.3 is the only recommended production ready version for edge side (standalone) deployment
+    image: apache/iotdb:0.13.3-node
     volumes:
-      - iotdb:/iotdb/data
+      - iotdb:/iotdb
     logging:
       driver: "json-file"
       options:
--- a/pom.xml
+++ b/pom.xml
@@ -158,7 +158,6 @@
 	<errorprone-annotations.version>2.2.0</errorprone-annotations.version>
 	<google-maps-services.version>0.10.0</google-maps-services.version>
 	<graalvm.js.version>22.3.1</graalvm.js.version>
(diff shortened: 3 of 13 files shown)


bossenti:
@SteveYurongSu Thank you so much for this awesome contribution 🙏🏼

tenthe:
Hi @SteveYurongSu, thank you so much for this PR and updating the IoTDB implementation.

I have a question regarding the "session". 
The session is created in the `onInvocation` and closed in the `onDetach` and what happens if the connection dies during the execution of the pipeline. Would it be possible to add a reconnection in the `onEvent` for this case or is this not necessary? 
Maybe this can be done when the "IoTDBConnectionException" occurs.

SteveYurongSu:
Hi @tenthe, thanks a lot for your reviewing. 

IoTDB has a native mechanism to handle the case that the connection dies. I will update the PR with the mechanism ASAP :-)

> Hi @SteveYurongSu, thank you so much for this PR and updating the IoTDB implementation.
> I tried it locally and everything worked fine. Thanks for updating the IoTDB version in the CLI.
> 
> I have a question regarding the "session". 
> The session is created in the `onInvocation` and closed in the `onDetach` and what happens if the connection dies during the execution of the pipeline. Would it be possible to add a reconnection in the `onEvent` for this case or is this not necessary? 
> Maybe this can be done when the "IoTDBConnectionException" occurs. 
>

tenthe:
Awsome, thanks a lot. 
I am excited to see how this will be implemented by IoTDB

SteveYurongSu:
Hi @bossenti, here is the update 😊

In streampipes-extensions/streampipes-sinks-databases-jvm/src/main/java/org/apache/streampipes/sinks/databases/jvm/iotdb/IotDb.java

I implemented the IoTDB sink by `SessionPool` instead of `Session`.

About `SessionPool`:
> We provide a connection pool (`SessionPool) for Native API. Using the interface, you need to define the pool size.
> 
> If you can not get a session connection in 60 seconds, there is a warning log but the program will hang.
> 
> If a session has finished an operation, it will be put back to the pool automatically. **If a session connection is broken, the session will be removed automatically and the pool will try to create a new session and redo the operation.**

tenthe (approved):

dominikriemer (approved):

dominikriemer:
Thank you @SteveYurongSu !

tenthe:
@SteveYurongSu thanks a lot for the super fast update.
I really like the new implementation

Issue #1158, opened 2023-01-25
Replace `consul-client` library with `consul-api` library


dominikriemer:
### Body

After upgrading Consul to the latest version, the currently used `consul-client` library throws an error related to the consul leader (explanation at [1]).

As this library isn't maintained anymore, we should switch to `consul-api` where this error doesn't occur.

[1] https://github.com/rickfast/consul-client/issues/457

### StreamPipes Committer

I acknowledge that I am a maintainer/committer of the Apache StreamPipes project.

PR #1169, opened 2023-01-26
Replace consul-client with consul-api library (#1158)

dominikriemer:
### Purpose

Replace consul library and add an initial version of improved handling of environment variables to facilitate easier testing.

### Remarks


dominikriemer:
To add consul integration tests, I had to refactor handling of environment variables so that these can be mocked.

Currently, we use the `Envs` enum to access env variables. The new version defines an `Environment` which provides variables and needs to be injected into the classes where env variable access is needed. Do you have feedback on this approach? 

I added this only for some variables required for service discovery and would extend the `Environment` class for all other variables if we decide that's a good approach.

tenthe (changes requested):
Hi @dominikriemer,
thanks for the changes and the update of the consul version.
I saw that the new version is missing in multiple places:
- [docker-compose](https://github.com/apache/streampipes/blob/dev/docker-compose.yml) in root directory
- Installer: All files in compose folder
- k8s folder

Do you know why there are changes in file: 3-getting-live-data-from-the-streampipes-data-stream is this an encoding problem?

Could there be breaking changes when users update their system to the newer version (e.g. variable value chagnes)?

Cheers,
Philipp

dominikriemer:
@tenthe Thanks for reviewing!
I've updated the consul versions in this branch directly.
Concerning the python file I think I committed this by accident.
There shouldn't be any variable changes right now, but we can have a separate discussion on providing variables currently stored in consul only as environment variables.

tenthe:
Great, thanks.
Do we also need to check if the Consul update itself works? So what happens to existing variables?

tenthe (approved):

dominikriemer:
I tried it locally and all variables where kept.
But it would be good to have some kind of automated tests for backwards compatibility between versions.

tenthe:
I will start a discussion on this because I am not sure yet how we can do this, or if we can even fully automate it.
I suggest that we discuss how and when we run the tests. Is this test part of the release process or could we even include it in the nightly test suite.
Once we know how to do it, it can be a `good first issue`

PR #1174, opened 2023-01-29
Sp 1157 upgrade third party services

dominikriemer:
### Purpose

Bump Influx to version 2.6, refactor code and add migrations to update existing installations.

### Remarks

* Bumps CouchDB to v3.3.1 (authentication required, but defaults are set)
* Bumps Influx to v2.6 (authentication required, defaults are set, migration flag required in setup files)
* Bumps Consul to 1.14.3 (no changes required)


docker-compose.yml
```diff
@@ -138,13 +141,20 @@ services:
...
       - INFLUXDB_REPORTING_DISABLED=false
       - INFLUXDB_ADMIN_ENABLED=true
+      - DOCKER_INFLUXDB_INIT_USERNAME=admin
+      - DOCKER_INFLUXDB_INIT_PASSWORD=sp-admin
+      - DOCKER_INFLUXDB_INIT_ADMIN_TOKEN=sp-admin
+      - DOCKER_INFLUXDB_INIT_ORG=sp
+      - DOCKER_INFLUXDB_INIT_BUCKET=sp
+      - DOCKER_INFLUXDB_INIT_MODE=setup
     volumes:
       - influxdb:/var/lib/influxdb
```
bossenti, line 156: Is the old one still required?
Is it used to ensure backwards compatibility?

installer/cli/.env
```diff
@@ -35,3 +35,8 @@ COMPOSE_PROJECT_NAME=streampipes
 # 
 # default: activemq
 #SP_MQTT_HOST=mosquitto
+
+
+# For database migration in v0.91.0 - set init mode to 'upgrade' to migrate an existing installation
+SP_INFLUX_INIT_MODE=setup
+#SP_INFLUX_INIT_MODE=upgrade
```
bossenti, line 42: We should provide an upgrade guide somewhere and also shortly describe this change in the PR's description (for the release notes)

installer/cli/deploy/standalone/couchdb/docker-compose.yml
```diff
@@ -16,9 +16,12 @@
 version: "3.4"
 services:
   couchdb:
-    image: couchdb:2.3.1
+    image: couchdb:3.3.1
```
bossenti, line 19: please include this in the PR's description as well

docker-compose.yml
```diff
@@ -138,13 +141,20 @@ services:
       spnet:
 
   influxdb:
-    image: fogsyio/influxdb:1.7
+    image: influxdb:2.6
     environment:
       - INFLUXDB_DATA_ENGINE=tsm1
       - INFLUXDB_REPORTING_DISABLED=false
       - INFLUXDB_ADMIN_ENABLED=true
+      - DOCKER_INFLUXDB_INIT_USERNAME=admin
```
bossenti, line 149: Do we want to align the naming style here with the envs of couchdb?

streampipes-commons/src/main/java/org/apache/streampipes/commons/constants/Envs.java
```diff
@@ -23,9 +23,9 @@ public enum Envs {
   SP_PORT("SP_PORT", null),
 
   @Deprecated(since = "0.90.0", forRemoval = true)
-  SP_CONSUL_LOCATION("CONSUL_LOCATION", "consul"),
+  SP_CONSUL_LOCATION("CONSUL_LOCATION", "consul", "localhost"),
```
bossenti, line 26: Is the deprecation here still valid?

dominikriemer: Yes, but we can remove this in the next version

docker-compose.yml
dominikriemer: These are the env variables that are supported by the official image, so we can't change these unfortunately

docker-compose.yml
dominikriemer: Yes, this is needed for automatic migration (we can't do a fully automated migration since the INIT_MODE must be set to either setup or upgrade).
We can remove the additional volume after the release

docker-compose.yml
bossenti: Makes sense 💡

tenthe (approved):

installer/cli/.env
dominikriemer: I created a ticket for this: #1206

PR #1180, opened 2023-01-30
[hotfix] Extend CustomRequestApi with new get request

tenthe:
### Purpose

I extended the class `CustomRequestApi` with a new method that takes additional query parameters for get requests to the API.

Here is a code sample:
```java
    Map<String, String> parameters = new HashMap<>();
    parameters.put("columns", "[id;COUNT;count_id]");

    var res = client
        .customRequest()
        .sendGet("api/v4/datalake/measurements/" + resourceId,
            parameters,
            SpQueryResult.class);

```

### Remarks


streampipes-client/src/main/java/org/apache/streampipes/client/util/StreamPipesApiPath.java
```diff
@@ -51,10 +57,30 @@ public StreamPipesApiPath addToPath(String pathItem) {
...
     pathItems.forEach(joiner::add);
-    return joiner.toString();
+    return appendQueryParameters(joiner.toString());
+  }
+
+  private String appendQueryParameters(String input) {
+
+    StringJoiner joiner = new StringJoiner("&");
+    for (Map.Entry<String, String> parameter : queryParameters.entrySet()) {
+      joiner.add(URLEncoder.encode(parameter.getKey(), StandardCharsets.UTF_8) + "="
```
dominikriemer, line 76: Can we extract the `URLEncoder.encode` operation to a separate method?

dominikriemer (approved):

PR #1181, opened 2023-01-30
Add output data streams to python functions

SvenO3:
### Purpose

This PR implements the support for output data streams for the python functions. This makes it possible to create a new data stream out of python and to write back live data to StreamPipes via that stream.
It also provides an implementation to use Online Machine Learning with the StreamPipesFunctions using the River library. You can now train an Online ML model on a data stream and sends the prediction back to StreamPipes with very little effort.

 ### Remarks


streampipes-client-python/streampipes_client/functions/machine_learning/__init__.py
```diff
@@ -0,0 +1,16 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
```
bossenti, line 2: lets move this into a dedicated module called `function_zoo`

bossenti (commented):
great work @SvenO3  🥇

bossenti:
almost forgot: please rebase the PR before we merge it because we upgraded `black`

streampipes-client-python/streampipes_client/functions/machine_learning/__init__.py
SvenO3: Should `function_zoo` be a submodule of the `functions` module. Or should it be seperated to differentiate between implementation and examples?

bossenti: I would see it as separate module on the same level like functions.
Should also be beneficial in terms of imports

bossenti:
Needs to be merged in sync with https://github.com/apache/streampipes/pull/1228

SvenO3:
> Needs to be merged in sync with #1228

What's the best way to do this?

bossenti:
Well, one of us has to do some work 😅
Either we merge #1228 first, then you need to update this PR accordingly or we do it the way around :)

Issue #880, opened 2022-12-22
Add git pre-commit hook for checkstyle


tenthe:
### Body

Since we have checkstyle enabled for all Java modules (see #820), it would be good to validate the styling with a Git pre-commit hook as well. 

### Mentoring
As this ticket is marked as good first issue: one of @tenthe, @RobertIndie  or @bossenti is happy to provide help for getting started, just tag (one of) them if you want to start working on this issue and need some help.

### StreamPipes Committer

- [X] I acknowledge that I am a maintainer/committer of the Apache StreamPipes project.

deexidee:
I could help with this one

bossenti:
That would be awesome @deexidee 🤩
Do you need anything to start with?

deexidee:
Sorry for missing, rough week again
Sure!
Actually, I never did anything related to git commit management. So I would be happy to see an example of a pre-commit hook.
And maybe searched poorly, but I couldn't find where pre-commit scripts are located. Are they in a different repo?

bossenti:
So, in generall a commit hook is a feature provided by git itself: https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks
You could for example write a bash file that should be executed when comitting.

For StreamPipes we have commit hooks at two places already present. The first one is for the UI codebase, it was contributed in #875 
The second one is used in our python module (`streampipes-client-python`) where we use the `pre-commit` library: https://pre-commit.com/

As far as I know this could be also used for Java with Checkstyle, but there are probably more Java-native solutions?
Maybe @dominikriemer @tenthe @smlabt @vesense @RobertIndie have an idea here

deexidee:
There is an ability to import a code style scheme from XML file in IntelliJ IDEA and Eclipse. Not sure if it's a suitable solution though

tenthe:
Yes, you can do this in your IDE. We also have a short description in our wiki how to do it [1].
You can find the checkstyle file in [2].
My suggestion would be to add checkstyle in your local development environment first. 
To test that everythinf is setup correctly, you can fix the checkstyle violations in #1192.

To ensure that there are no checkstyle violations we use a GitHub action. This is also the reason why the build in #1192 failed. To give the developer an even quicker feedback we want to integrate this check as a git pre-commit hook. That means a developer can only perform a git commit when there are no checkstyle violations.

I didn't do something like this before, but I guess we can also use the library https://pre-commit.com/ for that.

[...]

PR #1192, opened 2023-01-31
[SP-1102] - GET/v2/info/versions update and removement of deprecated endpoints 

deexidee:
### Purpose

Swagger doc description for endpoint "/v2/info/versions" at rest/impl/Version.class and removal of endpoints marked @Deprecated(since = "0.71.0", forRemoval = true).

### Breaking changes
Removal of deprecated API endpoints:
* `/v2/pipelines/own`
* `/v2/pipelines/system`
* `/v2/streams/own`
* `/v2/streams/own/<element_id>`

 ### Remarks


tenthe (commented):
Great, thank you very much!

tenthe:
They failed due to some checkstyle problems, see: https://github.com/apache/streampipes/actions/runs/4053261489/jobs/6981510149

Here is described how you can set this up checkstyle locally: https://cwiki.apache.org/confluence/display/STREAMPIPES/Code+Style+-+Java

There is also an issue on this topic, see #880. This might be a good next topic for you if you are interested. When we add the validation as a git pre-commit hook, we will avoid such situations in the future.

streampipes-rest/src/main/java/org/apache/streampipes/rest/impl/Version.java
```diff
@@ -33,6 +34,7 @@ public class Version extends AbstractAuthGuardedRestResource {
   @GET
   @Path("/versions")
```
bossenti, line 35: I would propose to rename the endpoint to `check`, `health-check` or `status`.
Any other ideas?

streampipes-rest/src/main/java/org/apache/streampipes/rest/impl/Version.java
```diff
@@ -33,6 +34,7 @@ public class Version extends AbstractAuthGuardedRestResource {
...
   @Produces(MediaType.APPLICATION_JSON)
+  @Operation(summary = "Provides health-check and information about current backend version.", tags = {"Version"})
```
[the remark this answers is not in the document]
bossenti: Alternatively, we could provide this endpoint twice: with and without authentication

bossenti:
Thanks a lot for your contribution @deexidee 🎉

streampipes-rest/src/main/java/org/apache/streampipes/rest/impl/Version.java
bossenti: @tenthe @dominikriemer

streampipes-rest/src/main/java/org/apache/streampipes/rest/impl/Version.java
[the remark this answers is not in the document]
dominikriemer: We also have an endpoint in the `Setup` resource which is unauthenticated and is used  by the UI to determine whether the core is running upon startup. We could move the version info there and can keep the authenticated endpoint for more detailed information. What do you think?

streampipes-rest/src/main/java/org/apache/streampipes/rest/impl/Version.java
dominikriemer: We could rename this to Status and add more information on registered extension services and other status info.
What info do we want to return here?

bossenti (approved):
Thanks a lot for your contribution @deexidee 🙏🏼 
We would be glad if you are interested in further contributions :)

PR #1286, opened 2023-02-14
Update data stream generator (#1258)

### LINKED ISSUE #1258: Simplify/restructure data stream generator
### Body

Currently, Python's data stream generator is a bit complicated and not very lucid.
We need to think how we can structure it better.
One solution could be to use `pydantic`s default functionalities. 

### StreamPipes Committer

SvenO3:
### Purpose

This PR resolves #1258

### Remarks

### DIFF
--- a/streampipes-client-python/streampipes/model/common.py
+++ b/streampipes-client-python/streampipes/model/common.py
@@ -87,19 +87,17 @@ class EventProperty(BasicModel):
     Data model of an `EventProperty` in compliance to the StreamPipes Backend.
     """
 
-    class_name: StrictStr = Field(
-        alias="@class", default_factory=lambda: "org.apache.streampipes.model.schema.EventPropertyPrimitive"
-    )
+    class_name: StrictStr = Field(alias="@class", default="org.apache.streampipes.model.schema.EventPropertyPrimitive")
     element_id: StrictStr = Field(default_factory=lambda: f"sp:eventproperty:{random_letters(6)}")
     label: Optional[StrictStr]
     description: Optional[StrictStr]
     runtime_name: StrictStr
-    required: StrictBool = False
-    domain_properties: List[StrictStr] = []
-    property_scope: StrictStr = "MEASUREMENT_PROPERTY"
-    index: StrictInt = 0
+    required: StrictBool = Field(default=False)
+    domain_properties: List[StrictStr] = Field(default_factory=list)
+    property_scope: StrictStr = Field(default="MEASUREMENT_PROPERTY")
+    index: StrictInt = Field(default=0)
     runtime_id: Optional[StrictStr]
-    runtime_type: StrictStr = "http://www.w3.org/2001/XMLSchema#string"
+    runtime_type: StrictStr = Field(default="http://www.w3.org/2001/XMLSchema#string")
     measurement_unit: Optional[StrictStr]
     value_specification: Optional[ValueSpecification]
 
@@ -132,7 +130,7 @@ class TopicDefinition(BasicModel):
     """
 
     class_name: Optional[StrictStr] = Field(
-        alias="@class", default_factory=lambda: "org.apache.streampipes.model.grounding.SimpleTopicDefinition"
+        alias="@class", default="org.apache.streampipes.model.grounding.SimpleTopicDefinition"
     )
     actual_topic_name: StrictStr = Field(default_factory=lambda: f"org.apache.streampipes.connect.{uuid4()}")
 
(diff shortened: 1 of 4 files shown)


streampipes-client-python/streampipes/model/common.py
```diff
@@ -73,17 +87,19 @@ class EventProperty(BasicModel):
     Data model of an `EventProperty` in compliance to the StreamPipes Backend.
     """
 
-    class_name: Optional[StrictStr] = Field(alias="@class")
-    element_id: Optional[StrictStr]
+    class_name: StrictStr = Field(
+        alias="@class", default_factory=lambda: "org.apache.streampipes.model.schema.EventPropertyPrimitive"
```
bossenti, line 91: In this case you could simply use `default` and provide the string value

streampipes-client-python/streampipes/model/common.py
```diff
@@ -73,17 +87,19 @@ class EventProperty(BasicModel):
...
+    )
+    element_id: StrictStr = Field(default_factory=lambda: f"sp:eventproperty:{random_letters(6)}")
     label: Optional[StrictStr]
     description: Optional[StrictStr]
     runtime_name: StrictStr
-    required: StrictBool
-    domain_properties: List[StrictStr]
-    property_scope: Optional[StrictStr]
-    index: StrictInt
+    required: StrictBool = False
```
bossenti, line 97: it would be beneficial to use the pydantic functionality here, so in this case `Field(default=False)`
Some applies for the upcoming rows, for an empty array you can use the `default_factory`

streampipes-client-python/streampipes/model/resource/data_stream.py
```diff
@@ -72,26 +73,31 @@ def convert_to_pandas_representation(self):
...
+    internally_managed: StrictBool = False
+    connected_to: Optional[List[StrictStr]]
+    event_grounding: EventGrounding = Field(default_factory=EventGrounding)
     event_schema: Optional[EventSchema]
-    measurement_capability: Optional[List[Optional[MeasurementCapability]]]
-    measurement_object: Optional[List[Optional[MeasurementObject]]]
-    index: Optional[StrictInt]
+    measurement_capability: Optional[List[MeasurementCapability]]
+    measurement_object: Optional[List[MeasurementObject]]
+    index: StrictInt = 0
```
bossenti, line 98: same as above

streampipes-client-python/streampipes/model/common.py
```diff
@@ -30,6 +33,17 @@
 ]
 
 
+def random_letters(n: int):
```
bossenti, line 36: I'm not really happy with this function, but don't really have a better idea 🤔 
I'll try to come with one

streampipes-client-python/streampipes/model/common.py
```diff
@@ -115,37 +131,41 @@ class TopicDefinition(BasicModel):
...
-    broker_hostname: StrictStr
-    topic_definition: TopicDefinition
-    port: StrictInt = Field(alias="kafkaPort")
+    class_name: StrictStr = Field(
+        alias="@class", default_factory=lambda: "org.apache.streampipes.model.grounding.NatsTransportProtocol"
+    )
+    element_id: StrictStr = Field(default_factory=lambda: f"sp:transportprotocol:{random_letters(6)}")
+    broker_hostname: StrictStr = "nats"
+    topic_definition: TopicDefinition = Field(default_factory=TopicDefinition)
+    port: StrictInt = Field(alias="kafkaPort", default_factory=lambda: 4222)
```
bossenti, line 151: Is `4222` the default port for all protocols?

streampipes-client-python/streampipes/model/common.py
SvenO3: Yes, this makes a lot more sense.

streampipes-client-python/streampipes/model/common.py
SvenO3: What don't you like about it?

streampipes-client-python/streampipes/model/common.py
SvenO3: I think it only for Nats, but the idea was to use Nats as default and you would need to configure this if you want to use other transport protocols.

streampipes-client-python/streampipes/model/common.py
bossenti: It's not about the implementation itself, it's more about that we only need it as default factory and does not really suit into the rest of the module 🤔

Issue #1289, opened 2023-02-15
Harmonize data set and data stream API


tenthe:
### Discussed in https://github.com/apache/streampipes/discussions/1115

<div type='discussions-op-text'>

<sup>Originally posted by **tenthe** January 17, 2023</sup>

## Harmonize `data set` and `data stream` APIs

We are currently looking at the Connect API and plan to refactor parts of it. Looking at the current implementation, I noticed that we have several cases that make the implementation more complex.

## Distinction between `data set` and `data stream` adapters

For example, we distinguish between `data set` and `data stream` adapters. Set adapters are treated as bounded streams, i.e. they stream a data set only once. Originally, this was added because it allows the user to replay existing events (e.g., from databases or files). However, I don't think this feature is used very often and we only have three implementations of set adapters. This feature adds a lot of overhead in many different places, such as the UI, the core, and extension services.

## Main features of current data sets

The main features of the data sets we currently use are:
- For the e2e tests to validate the processing elements
- Import a data set (e.g. CSV file) into the time-series storage

I think these are important and we should definitely keep them, but maybe we can find another solution to accomplish these tasks.

## Alternative solutions

New functionality:
- Add option to create adapters without starting them
- Add an option to the `FileStreamAdapter` to play the file only once

To import a dataset, a user (or the e2e tests) would need to create an adapter without starting it, create the pipeline, and then start the adapter.

## Recommendation

Since we don't have many benefits of the data set API, I would recommend removing it. This would also provide a clearer focus for StreamPipes because it focuses on streaming data produced by machines. Further, it will ease the implementation in many places without drawbacks in terms of functionality.
How do you see it?

PS: I would also like to harmonize the model for `GenericAdapters` and `SpecificAdapters`, but that is another discussion ;).

Cheers,
Philipp</div>

Issue #251, opened 2022-11-26
Create resource files for adapters


dominikriemer:
Remove old way of providing labels "Labels.from(internalId, labels, description)" to "Labels.withId(id)"

An example of the new version is shown in the ROS adapter [1].
 * To migrate to the new variant, first create a folder with the full package name in the resource directory. 
 * Then create a "strings.en" file
 ** The first two lines contain the adapter title and description
 ** Then provide the texts for all labels
 * Further add an icon (if not already provided) and create a documentation.md file with a detailed explanation of the functionality of the adapter

 

 

[1] [https://github.com/apache/incubator-streampipes-extensions/blob/dev/streampipes-connect-adapters/streampipes-connect-adapter/src/main/java/org/apache/streampipes/connect/adapters/ros/RosBridgeAdapter.java](https://github.com/apache/incubator-streampipes-extensions/blob/dev/streampipes-connect-adapters/streampipes-connect-adapter/src/main/java/org/apache/streampipes/connect/adapters/ros/RosBridgeAdapter.java)

Imported from Jira [STREAMPIPES-55](https://issues.apache.org/jira/browse/STREAMPIPES-55). Original Jira may contain additional context.
Reported by: zehnder.

bossenti:
Is mainly don

PR #1290, opened 2023-02-15
Remove set adapters

tenthe:
### Purpose

This PR is a **draft**. 
The PR removes the data set model as described in  #1289.

### Remarks


streampipes-extensions-management/src/main/java/org/apache/streampipes/extensions/management/connect/adapter/parser/util/JsonEventProperty.java
```diff
@@ -77,13 +79,15 @@ public static EventProperty getEventProperty(String key, Object o) {
...
+
     }
 
     if (resultProperty == null) {
       LOG.error("Property Type was not detected in JsonParser for the schema detection. "
-          + "This should never happen!");
+                + "This should never happen!");
     }
 
+    resultProperty.setDescription("");
```
bossenti, line 90: Can't we retrieve the description from somewhere or could we pass it to the method?

streampipes-extensions-management/src/main/java/org/apache/streampipes/extensions/management/connect/adapter/parser/CsvParser.java
```diff
@@ -0,0 +1,203 @@
...
+        .anyMatch("Header"::equals);
+
+    return new CsvParser(header, delimiter);
+  }
+
+
+  @Override
+  public ParserDescription declareDescription() {
+    return ParserDescriptionBuilder.create(ID, LABEL, DESCRIPTION)
+        .requiredTextParameter(Labels.from(DELIMITER, "Delimiter",
```
bossenti, line 96: This is a deprecated syntax. Is there any alternative or should we consider this a follow-up task?
I just want to avoid to increase the migration effort later

streampipes-extensions-api/src/main/java/org/apache/streampipes/extensions/api/extractor/IParameterExtractor.java
```diff
@@ -0,0 +1,119 @@
...
+  String secretValue(String internalName);
+
+  boolean slideToggleValue(String internalName);
+
+  String codeblockValue(String internalName);
+
+  String selectedColor(String internalName);
+
+  @Deprecated(since = "0.90.0", forRemoval = true)
+  String fileContentsAsString(String internalName) throws IOException;
```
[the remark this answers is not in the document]
dominikriemer: Yes, this is just the extracted interface which has all declarations from the existing implementation. Should we remove it now or in a later release?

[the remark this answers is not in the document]
bossenti: Is it still somewhen required, e.g., for migration?

streampipes-extensions-management/src/main/java/org/apache/streampipes/extensions/management/connect/adapter/parser/CsvParser.java
bossenti: Alright, https://github.com/apache/streampipes/issues/251 needs to be done first here, right?
Then let's handle this within 251 and try to integrate 251 within the 0.93.0 release

bossenti:
We should definitely think about writing documentation for this change as well.
From my point of view, we should at least cover:

- how to create an adapter with the new interfaces?
- how to implement additional parsers?

What do you think here? 
I would then create a follow-up issue

streampipes-extensions/streampipes-extensions-all-jvm/src/main/java/org/apache/streampipes/extensions/all/jvm/AllExtensionsInit.java
```diff
@@ -43,7 +43,7 @@ public SpServiceDefinition provideServiceDefinition() {
     return SpServiceDefinitionBuilder.create("org.apache.streampipes.extensions.all.jvm",
             "StreamPipes Extensions (JVM)",
             "", 8090)
-        //.merge(new ConnectAdapterInit().provideServiceDefinition())
+//        .merge(new ConnectAdapterInit().provideServiceDefinition())
```
bossenti, line 46: Can we remove this line of code?

bossenti:
@dominikriemer @tenthe I removed some data set artifacts in https://github.com/apache/streampipes/pull/1290/commits/b00c01c1c9a109aa12c94428d9d8c28434e52581
Can you please verify that this is correct?

streampipes-extensions/streampipes-connect-adapters-iiot/src/main/java/org/apache/streampipes/connect/iiot/adapters/plc4x/passive/Plc4xPassiveAdapter.java
```diff
@@ -18,100 +18,84 @@
 
 package org.apache.streampipes.connect.iiot.adapters.plc4x.passive;
 
-import org.apache.streampipes.extensions.api.connect.exception.AdapterException;
```
bossenti, line 21: What about this one?
Do we want to remove it entirely?

bossenti:
It would be great if we could port changes from https://github.com/apache/streampipes/pull/1638 to here

streampipes-extensions-api/src/main/java/org/apache/streampipes/extensions/api/extractor/IParameterExtractor.java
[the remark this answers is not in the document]
tenthe: I don't think it is necessary for migration. I think we can remove it.
We definitely need to test the migration extensively so we can determine if there will be a problem.
Should we create an issue for that?

streampipes-extensions-management/src/main/java/org/apache/streampipes/extensions/management/connect/adapter/parser/util/JsonEventProperty.java
tenthe: I am not quite sure. If I see it correctly the method `JsonEventProperty.getEventProperty` is mainly used in `GeoJsonParser`. There the description is not set.

streampipes-extensions-management/src/main/java/org/apache/streampipes/extensions/management/connect/adapter/parser/CsvParser.java
tenthe: Yes, I am not happy with the solution either. But we need to change the "labels" API in the future and I wasn't sure how else to provide the labels since the parsers don't have resources directly.

streampipes-extensions-api/src/main/java/org/apache/streampipes/extensions/api/extractor/IStaticPropertyExtractor.java
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ *
+ */
+
+package org.apache.streampipes.extensions.api.extractor;
+
+import org.apache.streampipes.model.graph.DataSinkInvocation;
+
+public interface IStaticPropertyExtractor extends IParameterExtractor<DataSinkInvocation> {
+}
```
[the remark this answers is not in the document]
dominikriemer: It's a bit clearer from an extension developer's point of view since no generics are used - but we should probably rename it to `IAdapterStaticPropertyExtractor` to be consistent with the other extractor in case the `IStaticPropertyExtractor` is only used for adapters...

streampipes-extensions-api/src/main/java/org/apache/streampipes/extensions/api/extractor/IParameterExtractor.java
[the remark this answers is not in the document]
bossenti: 👍🏼 
That would be great!

streampipes-extensions-api/src/main/java/org/apache/streampipes/extensions/api/extractor/IStaticPropertyExtractor.java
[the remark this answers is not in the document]
bossenti: Another PR is fine for - lets create an issue immediately so we won't forget about it

streampipes-extensions-management/src/main/java/org/apache/streampipes/extensions/management/connect/adapter/parser/util/JsonEventProperty.java
bossenti: Alright, we can also try to investigate this in a/the follow-up issue

streampipes-extensions/streampipes-connect-adapters-iiot/src/main/java/org/apache/streampipes/connect/iiot/adapters/plc4x/passive/Plc4xPassiveAdapter.java
bossenti: Done :)

bossenti:
> 

Here it is: https://github.com/apache/streampipes/issues/1645

bossenti:
> We should definitely think about writing documentation for this change as well. From my point of view, we should at least cover:
> 
> * how to create an adapter with the new interfaces?
> * how to implement additional parsers?
> * how to use the adapter migration?
> 
> What do you think here? I would then create a follow-up issue that we, in my opinion, should tackle within the 0.93.0 version

Here it is https://github.com/apache/streampipes/issues/1645

PR #1340, opened 2023-02-21
[#1291] Can't aggregate calculation results of the Math processing element

### LINKED ISSUE #1291: Can't aggregate calculation results of the Math processing element

Processing Elements

### What happened?

When calculating a new field with the  `Math` processing element, you can only aggregate the calculation result with `First`, `Last`, `Mode` and `Count` in the Data Explorer, but you can't use `Mean`, `Min`, `Max` or `Sum`. 

### How to reproduce?

Create a pipeline, use a data source with two numerical fields and use the  `Math` processing element to calculate a new field:
math_pipeline
When you agregate the calculation result with `First` every thing is fine:
aggregated_first
But if you try to show the `Mean` it does't work:
aggregated_mean

### Expected behavior

Maybe there's an error with the type of the calculation result.

_No response_

None

SteveYurongSu:
### Purpose

This fixes #1291.

**Root cause:**  
`InfluxStore` can not correctly storage `EventPropertyPrimitive` with `runtimeType = SO.NUMBER` as a number field in InfluxDB. 
A `EventPropertyPrimitive` with `runtimeType = SO.NUMBER` is stored directly as a `string` type field, which can not be used in some aggregation functions such as `mean`.

image
image

Actually, I have checked three possible points which may cause the issue:
1. Processor: wrong output field types of `MathOpProcessor `
2. DataLake: `InfluxStore` converts some field data types before insertion
3. Frontend: wrong number field selection logic in `FieldSelectionPanelComponent`

The check results:
1. fine
2. root cause
3. may have a potential issue: fields with type of `XSD.LONG` may not be shown by the panel. I fixed it.

### Tests
image
The field type of calculationResult is float now.

### Remarks

no

no

### DIFF
--- a/ui/src/app/data-explorer/services/data-explorer-field-provider-service.ts
+++ b/ui/src/app/data-explorer/services/data-explorer-field-provider-service.ts
@@ -194,10 +194,12 @@ export class DataExplorerFieldProviderService {
                 runtimeType === 'http://www.w3.org/2001/XMLSchema#float' ||
                 runtimeType === 'http://www.w3.org/2001/XMLSchema#double' ||
                 runtimeType === 'http://www.w3.org/2001/XMLSchema#integer' ||
+                runtimeType === 'http://www.w3.org/2001/XMLSchema#long' ||
                 runtimeType === 'https://schema.org/Number' ||
                 runtimeType === 'https://www.w3.org/2001/XMLSchema#float' ||
                 runtimeType === 'https://www.w3.org/2001/XMLSchema#double' ||
-                runtimeType === 'https://www.w3.org/2001/XMLSchema#integer'
+                runtimeType === 'https://www.w3.org/2001/XMLSchema#integer' ||
+                runtimeType === 'https://www.w3.org/2001/XMLSchema#long'
             );
         } else {
             return false;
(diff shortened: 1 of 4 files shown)


ui/src/app/data-explorer/services/data-explorer-field-provider-service.ts
```diff
@@ -194,10 +194,12 @@ export class DataExplorerFieldProviderService {
                 runtimeType === 'http://www.w3.org/2001/XMLSchema#float' ||
```
bossenti, line 194: I probably missed something here, but why do we have duplicate conditions here? line 193 until 197 is the same as 198 to 202, isn't it?

tenthe (approved):

tenthe:
Thank you for the PR and fixing the issue. 
The detailed description made the review quite easy.

ui/src/app/data-explorer/services/data-explorer-field-provider-service.ts
SteveYurongSu: The string prefix `https://www.w3.org/2001` appears only five times in all files, and it only appears in files related to the ui. It seems that we can remove the code associated with the prefix...

image

bossenti: Ah it is about `http` vs `https` 💡 
Didn't recognize the decisive difference so far 😅

SteveYurongSu: Haha, yes it's about http vs https 😂

But it's weird we have those https in our code, it seems to be useless... I have pushed a commit to remove them.

dominikriemer (approved):

PR #1361, opened 2023-02-28
Fix UI container cannot resolve backend IP correctly

RobertIndie:
### Purpose

Fix UI container cannot resolve backend IP correctly. We need to specify the k8s namespace in the backend domain.

This PR also ignore the file `installer/default.conf` which is generated automatically when deploying the cluster.

### Remarks


dominikriemer (approved):

dominikriemer:
Thank you @RobertIndie, I missed that!
Do you have any other ideas how we can improve the helm charts/which configurations are missing?

RobertIndie:
Hi @dominikriemer , Currently I am testing the helm deployment on my machine.  One thing I think we can improve is to expose configurations to choose which components should be deployed. 
Here is an example:
```yaml
components:
  ui: true
  core: true
  consul: true
  couchdb: false
...
```
I would like to investigate to try to improve the helm chart. :)

PR #1363, opened 2023-02-28
Add support for the Kafka broker in Python

dominikriemer:
### Purpose

This PR introduces support for the Kafka broker within StreamPipes Python.
With this a user is able to implement StreamPipes functions that interact with Kafka as a message broker.

### Remarks

### DIFF
--- a/streampipes-client-python/setup.py
+++ b/streampipes-client-python/setup.py
@@ -35,6 +35,7 @@ base_packages = [
     "pydantic>=1.10.2",
     "requests>=2.28.1",
     "nats-py>=2.2.0",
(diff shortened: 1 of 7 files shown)


tenthe (approved):

bossenti:
@dominikriemer @SvenO3 Would you mind to check the examples if we somewhere explicitly exclude other brokers than NATS?

SvenO3:
> @dominikriemer @SvenO3 Would you mind to check the examples if we somewhere explicitly exclude other brokers than NATS?

I can have a look at this.

PR #1408, opened 2023-03-11
Add search bar for Python docs

amaan-n-j:
### Purpose

Add a search bar to the Python docs that allows users to search the documents

### Remarks

### DIFF
--- a/streampipes-client-python/mkdocs.yml
+++ b/streampipes-client-python/mkdocs.yml
@@ -79,6 +79,10 @@ plugins:
       version_selector: true
       css_dir: css
       javascript_dir: js
+  - search:
+      lang:
+        - en
+      seperator: '[\s\-,:_!=\[\]()"/]+'
 
 extra_css:
   - stylesheets/extra.css


streampipes-client-python/mkdocs.yml
```diff
@@ -79,6 +79,10 @@ plugins:
       version_selector: true
       css_dir: css
       javascript_dir: js
+  - search:
+      lang:
+        - en
+      seperator: '[\s\-,:_!=\[\]()"/]+'
```
bossenti, line 85: Why did you only special chars as separator?
We are on version 8 of mkdocs-material so the default seperator is `'[\s\-\.]+'`
What's your idea about changing this? :)

bossenti:
@amaan-n-j thanks for your contribution 🙏🏼

streampipes-client-python/mkdocs.yml
amaan-n-j: I thought user might not know the exact thing to search, may end up adding some special character. If you want to i can remove them and keep the defaults.

bossenti: Sounds valid
Let's try it like this

bossenti (approved):

Issue #1564, opened 2023-05-09
Add feature to assign labels to StreamPipes resources


dominikriemer:
### Body

In larger setups with many resources such as pipelines or data explorer views, it might be hard to find a specific resource. We can add a new feature to assign "labels" (similar to Github labels) to these resources so that resources can be more easily discovered and filtered.

### StreamPipes Committer

I acknowledge that I am a maintainer/committer of the Apache StreamPipes project.

PR #1568, opened 2023-05-12
Add configuration view to create labels (#1564)

dominikriemer:
### Purpose

Add a new page in the configuration section to create and edit labels, which will be later used to better organize StreamPipes resources.

### Remarks


tenthe (approved):

bossenti (commented):
@dominikriemer I have one follow-up question: Currently, it would be possible to create two labels with the same name, right?

PR #1628, opened 2023-05-28
chore: add missing support of NATS as messaging protocol

bossenti:
### Purpose

Adds missing support for NATS as messaging protocol.
Relates to #1584.

### Remarks


bossenti:
Is it sufficient to simply add NATS?
Or are there any other aspects we should account for?
Is there still something missing?

dominikriemer (approved):

dominikriemer:
@bossenti looks good to me!
I don't think anything else is needed.

tenthe (approved):

PR #1641, opened 2023-05-30
[hotfix] Fix bug in time selection of data explorer cypress tests

dominikriemer:
### Purpose

Some data explorer cypress e2e tests used a static date value (May 30th) in the time selection widget, causing some tests to fail from today on ;-)
The PR now uses a future date and properly clicks on the time selection fields in the data explorer.

### Remarks


bossenti (approved):
well spotted 👍🏼

PR #1653, opened 2023-06-04
add validator that allows to pass list of columns to data lake query (#1374)

### LINKED ISSUE #1374: Convenient `columns` query parameter for data lake measure
### Body

Currently, the python client expects a comma separated string when specifying the columns to be returned ( query parameter `columns`, see [here](https://github.com/apache/streampipes/blob/dev/streampipes-client-python/streampipes/endpoint/api/data_lake_measure.py#LL42)) for a data lake measure. This is quite inconvenient and should be replaced by the option to provide a list of strings which is then converted to a comma separated string for the actual query param. This transformation should be done with a [pydantic validator](https://docs.pydantic.dev/usage/validators/).

### Mentoring
As this issue is marked as good first issue: one of @SvenO3 , or @bossenti are happy to provide help for getting started, just tag (one of) them if you want to start working on this issue and need some help.
This issue is ideal for people that are interested in getting started with StreamPipes and making their first steps in our Python library.

### StreamPipes Committer

jjbiggins:
### Purpose

Added pydantic validator for pre columns, and joins a list into a comma delimited string.

Resolves #1374  

### Remarks

PR introduces (a) breaking change(s): 
no

PR introduces (a) deprecation(s): 
no

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -59,7 +59,7 @@
         "quill": "1.3.7",
         "roboto-fontface": "0.10.0",
         "rxjs": "^7.8.0",
-        "shepherd.js": "^11.0.1",
+        "shepherd.js": "^11.1.1",
         "showdown": "1.9.1",
         "stream-browserify": "3.0.0",
         "swagger-ui": "4.18.0",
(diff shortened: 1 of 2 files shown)


bossenti (commented):
Thanks a lot for your contribution @jjbiggins 🙏🏼 

I've left a small remark in the source code.
Additionally, it would be great if you could remove the old regex to check the comma separated string in [line 82](https://github.com/apache/streampipes/pull/1653/files#diff-501eb1f805a04e5126144a118b9c1273c59863d60310a153c213d6f4dd5e5982L82).
Finally, bringing recent changes from `dev` branch back to this branch makes the PR ready to merge 🙂

jjbiggins:
No, problems. I got this done, but before I commit. I just want to clarify the change involving the removal of the regex string. 

That regex string is being used to validate the entire final string; both that it is comma delimited, and also contains only valid characters.

The original validator I wrote is a pre-validator, and so it still uses this regex. If I remove the regex, the whitespace character test fails.

The regex limited the valid characters for a column to alphanumeric and underscores. If I remove it, that validation will no longer happen. Someone could pass a list as such: ["tes-", " 123"]; this would no long raise an Exception here. 

I just wanted to raise this point. If that's already known or handled elsewhere, I will commit the changes as you asked.

Thanks

Issue #774, opened 2022-11-26
Python client inegration test


dominikriemer:
Write a basic integration test for the Python client.
Should be covering at least the result of STREAMPIPES-608

Imported from Jira [STREAMPIPES-609](https://issues.apache.org/jira/browse/STREAMPIPES-609). Original Jira may contain additional context.
Reported by: bossenti.

PR #1758, opened 2023-07-10
[hotfix] Improve handling of data lake measures in API

dominikriemer:
### Purpose

Improve the handling of data lake measures when using the API:
* Keep document ID if present
* Return 404 if the requested measure is not present

### Remarks


bossenti:
@dominikriemer is there a need to account for this changes in the Python client?

tenthe (approved):

dominikriemer:
> @dominikriemer is there a need to account for this changes in the Python client?

I had a brief look and I don't think so - can you please double-check if it's ok? The Java client had the issue that it tried to deserialize the empty response.

bossenti:
> > @dominikriemer is there a need to account for this changes in the Python client?
> 
> 
> 
> I had a brief look and I don't think so - can you please double-check if it's ok? The Java client had the issue that it tried to deserialize the empty response.

Should be fine I guess :) 

@tenthe time to work together on #774 😉

Issue #1769, opened 2023-07-12
The dashboard fails to load the element whose name contiains '/'


RobertIndie:
### Apache StreamPipes version

dev (current development state)

### Affected StreamPipes components

Backend

### What happened?

The dashboard fails to load the element name contiains '/'

### How to reproduce?

- Creating a pipeline that contains a data lake named 'test/001'
- Create a dashboard and add a widget for the data lake
- The dashboard couldn't find the widget
<img width="859" alt="image" src="https://github.com/apache/streampipes/assets/16974619/569e5a86-9ccf-44a7-9b6d-df5233f7411d">

### Expected behavior

The dashboard could find the widget and display.

### Additional technical information

The root cause is that the URL is corrupted with '/'. Here is the request URL for this widget:
```
http://localhost:80/streampipes-backend/api/v3/datalake/pipelines/1cbeb7f2e1da494a9b9f45e36f63e513/test/001
```

We need to use the escape character here. Or reject the use of the name containing '/' when creating the element.

### Are you willing to submit a PR?

None

dominikriemer:
@RobertIndie good catch!
I added a method to replace forbidden characters with an underscore, similar as we are rewriting fields which contain reserved keywords.
(I first tried to configure Tomcat to allow URL-encoded slashs, but couldn't manage to get the configuration property working, so I went for this solution.)

PR #1774, opened 2023-07-15
Replace forbidden keywords in data lake measure (#1769)

dominikriemer:
### Purpose

See #1769 

### Remarks


streampipes-data-explorer-commons/src/main/java/org/apache/streampipes/dataexplorer/commons/sanitizer/MeasureNameSanitizer.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package org.apache.streampipes.dataexplorer.commons.sanitizer;
+
+import java.util.List;
+
+public class MeasureNameSanitizer {
+
+  private final List<String> forbiddenChars = List.of(
+      "/"
```
bossenti, line 26: Didn't we also need to exclude `-`? 🤔 (cc @tenthe)
We also might need to replace other characters like `&`, did you already check that?

tenthe: Hey,

I have tested several special characters. Here are the results:
grafik

It seems that `?`, `"`, `=` should also be escaped.
Cheers, 
Philipp

dominikriemer: Thanks @tenthe for providing the tests!

tenthe (approved):

PR #1800, opened 2023-07-27
fix wrong license headers in streampipes-pipeline-elements-experimental module

bossenti:
### Purpose

This PR removes wrongly introduced license headers.

- [ ] adapt checkstyle configuration so that it no longer complains about missing Apache headers

### Remarks


pom.xml
```diff
@@ -1720,7 +1720,7 @@
 
                         <!-- External files -->
                         <exclude>
-                            streampipes-extensions/streampipes-processors-transformation-flink/src/main/java/com/kohlschutter/boilerpipe/**
+							streampipes-extensions/streampipes-pipeline-elements-experimental-flink/src/main/java/com/kohlschutter/boilerpipe/**
```
bossenti, line 1723: This exclude statement needs to be adapted

PR #1835, opened 2023-08-20
Bump Kafka version to 3.x

dominikriemer:
### Purpose

Bump Kafka from 2.2 to 3.5 and enable KRaft mode without Zookeeper. 

### Remarks

The docker-compose file is updated which should be mentioned in the migration guide since it's not enough to simply bump the StreamPipes version in the env file.


tenthe (approved):

bossenti:
Do we really want to drop support for Zookeeper directly?
Would be easier for existing deployments to still use zookeeper.

In addition:
> The docker-compose file is updated which should be mentioned in the migration guide since it's not enough to simply bump the StreamPipes version in the env file.

How would the migration look like?

dominikriemer:
> Do we really want to drop support for Zookeeper directly? Would be easier for existing deployments to still use zookeeper.
> 

@bossenti What do you mean by existing deployments? Users using the bundled Kafka or users who provide their own Kafka infrastructure?

> In addition:
> 
> > The docker-compose file is updated which should be mentioned in the migration guide since it's not enough to simply bump the StreamPipes version in the env file.
> 
> How would the migration look like?

It's not really a migration, but users will need to either use the latest docker-compose file or copy the new environment variables from the docker compose file to their existing file.

bossenti:
>or copy the new environment variables from the docker compose file to their existing file

Okay, so it is still possible to run StreamPipes with kafka and zookeeper if required?

RobertIndie (approved):

RobertIndie:
> Okay, so it is still possible to run StreamPipes with kafka and zookeeper if required?

Yes, I think so.

Issue #1688, opened 2023-06-15
New Processor: Round Numeric Values


tenthe:
### Description
A user should have the ability round numeric values in a pipeline. Therefore, a new processor is required that provides the functionality to round numeric numbers of the event.

### Example Configuration
- **Multiselect**: The processor should allow the user to select all numeric values in the data stream for rounding.
- **Numeric Input**: The user should be able to specify the precision to which the values should be rounded.

### Output Schema
The input schema should not be changed

### Additional Functionality
If you have any other ideas for the functionality of this new processor, please feel free to add them here.

### Mentoring
As this issue is marked as good first issue: one of @dominikriemer, @tenthe, or @bossenti are happy to provide help for getting started, just tag (one of) them if you want to start working on this issue and need some help.
This issue is ideal for people that are interested in getting started with StreamPipes and its SDK.

### StreamPipes Committer

I acknowledge that I am a maintainer/committer of the Apache StreamPipes project.

muyangye:
Hi, I have been using Streampipes for a while and I absolutely love the project. Since no one has taken this issue yet I can try. Any suggestions to get started? Like is there a guide to create a new processor and should this round processor be under math or filter or transformation? Thanks!

bossenti:
Thanks a lot for your nice words @muyangye!
We would really appreciate a contribution here.

You can find a tutorial on implementing processing elements here: https://streampipes.apache.org/docs/extend-tutorial-data-processors
I would simply assign this element to the math transformation elements

Feel free to reach out, whenever you face any issue. We are happy to guide you 🙂

muyangye:
I have finished the implementation and tested locally using a FileStream source, should I submit a PR now (sorry if this is a dumb question, this is my first time contributing to OSS)? That being said, I still have 4 questions (and my current approach):

- I am using Labels.from() to create a label and according to source code this method is marked as deprecated. Is this a problem?
- Just want to confirm the exact meaning of rounding, currently I am using the conventional round half up.
- When I try to update the new processor, there's an error message "Icon missing", is there a place to add an icon to the processor?
- This question is not really related to this processor but I am curious: Why only event.addField()'s implementation uses runtime name (what's after s0::) and other functions such as getFieldBySelector() and removeFieldBySelector() use the selector?

Cheers,
Muyang

dominikriemer:
Hi @muyangye and thanks for the PR!
Here are some answers to your questions:
[...]

PR #1858, opened 2023-08-29
[#1688] implement new round processor

muyangye:
### Purpose
See #1688 

### Remarks


muyangye:
Hi I just saw these errors, I tried to run "mvn clean install" but this error popped up first so I didn't wait until it finishes.
error
I just looked into the errors on Github CI which are the same as mine but it kept running and showed the checkstyle errors. I have pushed an update that should fix them. :-) @bossenti

dominikriemer:
Hi @muyangye the errors appears when using Java 11 - Java 17 is the currently supported minimum version so that the error should disappear when you are building with java 17.

muyangye:
Hi @dominikriemer , yep I updated to Java 17 and all tests passed locally before pushing the new commit. Good to see Github CI also passed.

streampipes-extensions/streampipes-processors-transformation-jvm/src/main/java/org/apache/streampipes/processors/transformation/jvm/processor/round/RoundProcessor.java
```diff
@@ -0,0 +1,85 @@
...
+                           EventProcessorRuntimeContext runtimeContext) throws SpRuntimeException {
+    fieldsToBeRounded = parameters.extractor().mappingPropertyValues(FIELDS);
+    numDigits = parameters.extractor().singleValueParameter(NUM_DIGITS, Integer.class);
+  }
+
+  @Override
+  public void onEvent(Event event, SpOutputCollector collector) throws SpRuntimeException {
+    for (String fieldToBeRounded : fieldsToBeRounded) {
+      double value = event.getFieldBySelector(fieldToBeRounded).getAsPrimitive().getAsDouble();
+      double roundedValue = BigDecimal.valueOf(value).setScale(numDigits, RoundingMode.HALF_UP).doubleValue();
```
bossenti, line 76: Could you please make the rounding mode a user configurable input too?

bossenti (commented):
Thanks a lot @muyangye for your contribution 🙏🏼 
The PR already looks pretty good.

In addition to my inline comment, I'm afraid that we can't use the icon you provided (although I really like it). We don't take icons from external sources to prevent any copyright issue. Would be great if you could come with an alternative.
If you don't want to, I can also take over

streampipes-extensions/streampipes-processors-transformation-jvm/src/main/java/org/apache/streampipes/processors/transformation/jvm/processor/round/RoundProcessor.java
muyangye: For sure, I will support all the rounding modes specified [here](https://docs.oracle.com/javase/8/docs/api/java/math/RoundingMode.html). Should we also support custom number to round up (e.g. 3 to round up means 1.3 will be rounded to 2)? I believe this is somewhat unnecessary because the user can just apply some mathematical transformations to achieve this. What do you think?

muyangye: For the icon, I am really really really bad at hand/digital drawing... 🤣 It would be great if you can help come up with the icon.

bossenti: > Should we also support custom number to round up (e.g. 3 to round up means 1.3 will be rounded to 2)? 
Sorry, I don't really get your point. Can you please rephrase it? 

No, worries. I'll come up with one 🙂

muyangye: > > Should we also support custom number to round up (e.g. 3 to round up means 1.3 will be rounded to 2)?
> 
> Sorry, I don't really get your point. Can you please rephrase it?
> 
> > For the icon, I am really really really bad at hand/digital drawing... 🤣 It would be great if you can help come up with the icon.
> 
> No, worries. I'll come up with one 🙂

Sorry for my confusing phrasing. I basically mean I believe only [these modes](https://docs.oracle.com/javase/8/docs/api/java/math/RoundingMode.html) need to be supported. Is there any other rounding mode that you think users will want to use?

bossenti: Ah yes, I think they are perfectly fine 👍🏼

bossenti (approved):

tenthe:
Hey @muyangye, 
thanks a lot for providing the PR

PR #1905, opened 2023-09-10
Reordering Parameters in the PostgreSQL Database Sink Connector

fabianmerino:
### Purpose

This PR resolves an issue found in onInvocation() function, responsible for creating the parameters to establish the connection to the PostgreSQL database in the project. The error occurred when parameters were provided in the incorrect order, leading to connection failures.

### Remarks

Reordered the parameters to ensure the correct sequence is followed.

The correct sequence is: https://github.com/apache/streampipes/blob/dev/streampipes-extensions/streampipes-sinks-databases-jvm/src/main/java/org/apache/streampipes/sinks/databases/jvm/postgresql/PostgreSqlParameters.java#L26

### DIFF
--- a/streampipes-extensions/streampipes-sinks-databases-jvm/src/main/java/org/apache/streampipes/sinks/databases/jvm/postgresql/PostgreSqlSink.java
+++ b/streampipes-extensions/streampipes-sinks-databases-jvm/src/main/java/org/apache/streampipes/sinks/databases/jvm/postgresql/PostgreSqlSink.java
@@ -87,9 +87,9 @@ public class PostgreSqlSink extends StreamPipesDataSink {
         hostname,
         port,
         dbName,
-        tableName,
         user,
-        password,
+        password,      
+        tableName,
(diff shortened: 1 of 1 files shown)


dominikriemer:
Hi @fabianmerino thanks for the PR!
When comparing the signature of the `PostgreSqlParameters` class,:
*  `DataSinkInvocation graph, String postgreSqlHost, Integer postgreSqlPort, String databaseName, String tableName, String user, String password, Boolean sslEnabled`

with the call in the `PostgreSqlSink`:
* `parameters.getModel(), hostname, port, dbName, tableName, user, password, sslSelection.equals(SSL_ENABLED)`

this seems to match? Am I missing something?

fabianmerino:
It's my mistake, I was trying to see where this error comes from:

src1


image


And I didn't check properly, sorry. Instead I'll open an issue.

bossenti:
@fabianmerino thanks for the PR anyways!

Would be great to open an issue so we can track & investigate this error

PR #1914, opened 2023-09-11
Add documentation for S7 and Modbus adapters

dominikriemer:
### Purpose

Add documentation for the PLC4X S7 and Modbus adapters.

### Remarks

### DIFF
--- a/streampipes-extensions/streampipes-connect-adapters-iiot/src/main/resources/org.apache.streampipes.connect.iiot.adapters.plc4x.modbus/documentation.md
+++ b/streampipes-extensions/streampipes-connect-adapters-iiot/src/main/resources/org.apache.streampipes.connect.iiot.adapters.plc4x.modbus/documentation.md
@@ -16,7 +16,7 @@
   ~
   -->
 
-## ISS Adapter
+## Modbus
 
 <p align="center"> 
     <img src="icon.png" width="150px;" class="pe-image-documentation"/>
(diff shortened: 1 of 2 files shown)


streampipes-extensions/streampipes-connect-adapters-iiot/src/main/resources/org.apache.streampipes.connect.iiot.adapters.plc4x.s7/documentation.md
```diff
@@ -26,8 +26,62 @@
...
+  datatype. The `Runtime Name` will be the StreamPipes-internal name of the field, which will also show up in the data
+  explorer and pipeline editor. The `Node Name` refers to the node address of the PLC, e.g., `%Q0.4`. Finally, the data
+  type can be selected from the available selection. Currently available data types
+  are `Bool`, `Byte`, `Int`, `Word`, `Real`, `Char`, `String`, `Date`, `Time of Day` and `Date and Time`.
+* Instead of providing the node information manually, a CSV file can be uploaded. The CSV file can, for instance, be
+  exported from TIA and then be enriched with the appropriate runtime names. This is especially useful when many fields
+  should be added as nodes. Here is an example export enriched with the runtime name:
+
+```
+Runtime Name,Path,Data Type,Node Name,Comment,Hmi Visible,Hmi Accessible,Hmi Writeable,Typeobject ID,Version ID
```
tenthe, line 76: Only the `Runtime Name`, `Data Type`, and `Node Name` columns are required. The mapping is done based on the title of the corresponding`StaticProperty`. All other fields are ignored. I'd suggest removing the other columns for simplicity. Maybe we can add a comment if there are more columns it is not a problem because they will be ignored.

streampipes-extensions/streampipes-connect-adapters-iiot/src/main/resources/org.apache.streampipes.connect.iiot.adapters.plc4x.modbus/documentation.md
```diff
@@ -26,8 +26,34 @@
...
+The PLC port refers to the port of the PLC, such as 502.
+
+### Node ID
+
+The Node ID refers to the ID of the specific device.
+
+### Nodes
+
+The `Nodes` section requires configuration options for the individual nodes.
+Nodes can be either imported from a comma-separated CSV file, or can be directly assigned in the configuration menu.
```
tenthe, line 52: Maybe we can here also provide a small sample CSV, that a user can copy to get started.

tenthe (changes requested):
Thank you very much for writing the documentation for the adapters. I find it very helpful. I added two small comments regarding the CSV files. 
I think it would be very helpful to have good CSV examples here, because then we always have a good reference that we can copy and show to others.

streampipes-extensions/streampipes-connect-adapters-iiot/src/main/resources/org.apache.streampipes.connect.iiot.adapters.plc4x.s7/documentation.md
dominikriemer: I left one additional field so that it is clear that the CSV  does not exactly need to contain the required fields.

streampipes-extensions/streampipes-connect-adapters-iiot/src/main/resources/org.apache.streampipes.connect.iiot.adapters.plc4x.modbus/documentation.md
dominikriemer: Added a CSV example.

tenthe (approved):

PR #1918, opened 2023-09-12
Improve and centralize HTTP requests to extensions

dominikriemer:
### Purpose

Improve handling of HTTP requests from core to extensions and add auth tokens to all requests (in case extensions are running in auth mode).

### Remarks


streampipes-connect-management/src/main/java/org/apache/streampipes/connect/management/management/GuessManagement.java
```diff
@@ -57,10 +56,8 @@ public GuessSchema guessSchema(AdapterDescription adapterDescription)
     var objectMapper = JacksonSerializer.getObjectMapper();
     var description = objectMapper.writeValueAsString(adapterDescription);
     logger.info("Guess schema at: " + workerUrl);
-    Response requestResponse = Request.Post(workerUrl)
-        .bodyString(description, ContentType.APPLICATION_JSON)
-        .connectTimeout(1000)
-        .socketTimeout(100000)
+    Response requestResponse = ExtensionServiceExecutions
+        .extServicePostRequest(workerUrl, null, description)
```
tenthe, line 60: Is it possible that this `null` value causes an exception? Because in the `AuthTokenUtils.getAuthToken` the `ResourceId` is checked for null and then throws an exception

tenthe (approved):

.github/workflows/pr-validation.yml
```diff
@@ -100,6 +100,11 @@ jobs:
         working-directory: ./installer/compose
         run: sleep 70
 
+      - name: Print debug output
```
tenthe, line 103: This can be removed again, right?

PR #1945, opened 2023-09-18
Add configuration hint for Kafka users.

obermeier:
### Purpose
I think we have very nice examples of how to use the StreamPipes Python client. I have a suggestion on how to improve the user experience further related to Kafka configuration for less experienced users .

In the given example, localhost as hostname is used as the hostname. I think this could lead to the assumption that arbitrary hostnames or IPs are possible by default.
For this reason, I suggest adding a small hint for non-Kafka-experienced users that the external address should be advertised by Kafka by adjusting KAFKA_ADVERTISED_LISTENERS.

  Example for the docker-compose env:   ` KAFKA_ADVERTISED_LISTENERS: PLAINTEXT://:9092,OUTSIDE://{EXTERNAL_ADDRESS}:9094`

"What do you think of this idea? 
Do you prefer another place to explain situations like this?"

PR introduces (a) breaking change(s): <no>

PR introduces (a) deprecation(s): <no>


tenthe:
Hi @obermeier,
thanks for the PR. I really like the idea.
I also think it makes sense to write a small comment in the docker-compose file. So that a user knows directly where and how to change the kafka configurations.
Cheers,
Philipp

bossenti:
this is a great addition @obermeier 🙏🏼

dominikriemer (approved):

tenthe (approved):

PR #1955, opened 2023-09-22
Introduce Quickstart deployment mode

Cccccrystall:
### Purpose
This Pull Request introduces a new docker-compose.quickstart.yml file and related configurations, aiming to streamline and enhance the user experience for quick project setups. Here is a comprehensive summary of the changes and additions:
 
**New Docker Compose File**
 
Introduced docker-compose.quickstart.yml, derived from the existing docker-compose.full.yml.
Added additional configuration for the script-runner service, detailing build context, volume mapping (including the mapping of zip_folder), network dependencies, and service dependencies.
 
**Dockerfile for Script-Runner**
 
Created a Dockerfile specifying the base image as ubuntu:latest.
Installed necessary utilities: curl and jq.
Added and made executable two Bash scripts: upload.sh and wait-for-it.sh.
 
**Upload Script**
 
The script checks environment variables for user credentials, logs in to obtain an access token, and processes and uploads ZIP configuration files located in /zip_folder to the backend service.
 
**ZIP Configuration Files and Folder Mapping**
 
Included a new folder zip_folder containing ZIP configuration files to be used by the upload.sh script.
The zip_folder is mapped to the script-runner container, allowing direct access and processing of the ZIP files within the container.
The introduction of zip_folder mapping facilitates easy additions of default ZIP configurations. Whenever there is a need to introduce new default configurations, we can simply add the corresponding ZIP files to the zip_folder, making the process more efficient and user-friendly.
 
**Wait-for-it Script**
 
Incorporated the wait-for-it.sh script, enhancing the container’s capability to wait for the availability of the dependent backend service before executing upload.sh.
Usage Instructions:
 
To utilize the new configurations, the user should build the script-runner service with the command: docker-compose -f docker-compose.quickstart.yml build script-runner.
Subsequently, to trigger the setup, the user should run: docker-compose -f docker-compose.quickstart.yml up -d.
 
**Future Plan**
As part of our ongoing efforts to enhance user experience, we plan to package the script-runner service as a separate Docker image and upload it to Docker Hub. This step will allow users to skip the build step and trigger the service directly, thereby further simplifying the setup process.
 
These enhancements are designed to facilitate easier and more efficient project setups, especially beneficial for users seeking a quickstart option. By introducing the script-runner service with automated script executions and additional configurations, we are enabling seamless interactions and operations for different project use cases.

**Quick Demo**
output

### Remarks


Cccccrystall:
Hi @bossenti @RobertIndie @tenthe @dominikriemer, I kindly request your assistance in reviewing this Pull Request. Feel free to add your comments or suggestions; your input will be highly appreciated. Thank you in advance.

installer/compose/docker-compose.quickstart.yml
RobertIndie, line 1: It seems better to be based on https://github.com/apache/streampipes/blob/dev/installer/compose/docker-compose.yml. This seems more lightweight.

installer/compose/upload.sh
```diff
@@ -0,0 +1,64 @@
...
+
+# Check if SP_INITIAL_ADMIN_PASSWORD is empty; if so, set PASSWORD to a default value
+if [ -z "$SP_INITIAL_ADMIN_PASSWORD" ]; then
+  PASSWORD="admin"
+else
+  PASSWORD="$SP_INITIAL_ADMIN_PASSWORD"
+fi
+
+echo $USERNAME
+echo $PASSWORD
```
RobertIndie, line 35: We don't need to echo them.

dominikriemer:
@Cccccrystall I think it's better to put the Dockerfile, zip files and scripts into a separate folder so that users don't confuse the Dockerfile with the project's other Dockerfiles.

Also, I notice there are three zip files..these seem to be more placeholders with very basic examples - do you plan to provide a full example that helps users doing their first steps?

dominikriemer:
The checks are currently faliing due to missing license headers.

Cccccrystall:
> @Cccccrystall I think it's better to put the Dockerfile, zip files and scripts into a separate folder so that users don't confuse the Dockerfile with the project's other Dockerfiles.
> 
> Also, I notice there are three zip files..these seem to be more placeholders with very basic examples - do you plan to provide a full example that helps users doing their first steps?

Hello, @dominikriemer! I greatly appreciate your valuable suggestions! They indeed have been quite beneficial. In an effort to enhance the user experience, I'm keen on understanding which type of asset would be most suitable for users trying Streampipes for the first time. However, in my own search efforts across platforms like Youtube, I seem to only encounter basic examples for tutorials. I'm eager to provide a more comprehensive example for our users, given your expertise, @dominikriemer, would you be able to provide some guidance on finding the most suitable asset? Your insights would be greatly beneficial in helping to enhance the initial user experience with Streampipes Quickstart mode.

installer/compose/docker-compose.quickstart.yml
Cccccrystall: Hi, @RobertIndie!  I will change this when I submit new code. Thanks for the advice!

Cccccrystall:
Hi @dominikriemer, @bossenti! Two weeks ago I made some changes, including providing new example assets, merging quickstart scripts and resources into a separate folder, and changing quickstart.yml based on compose.yml. Could you review these changes for me and give me some advice? I would appreciate it!

installer/compose/README.md
```diff
@@ -90,6 +91,18 @@ docker-compose -f docker-compose.full.yml down
...
+We introduce quickstart deployment that comes with a set of predefined StreamPipes assets, to use the quickstart mode, just build the Docker image and start **quickstart** option:
+```bash
+docker-compose -f docker-compose.quickstart.yml build script-runner
+docker-compose -f docker-compose.quickstart.yml up -d
+# go to after all services are started http://localhost
+```
+Stopping the **quickstart** option:
+```bash
+docker-compose -f docker-compose.quickstart.yml down
+#docker-compose -f docker-compose.quickstart.yml down
```
bossenti, line 103: This comment is actually a duplicate right?

installer/compose/README.md
```diff
@@ -90,6 +91,18 @@ docker-compose -f docker-compose.full.yml down
 #docker-compose -f docker-compose.full.yml down
 ```
 
+We introduce quickstart deployment that comes with a set of predefined StreamPipes assets, to use the quickstart mode, just build the Docker image and start **quickstart** option:
+```bash
+docker-compose -f docker-compose.quickstart.yml build script-runner
+docker-compose -f docker-compose.quickstart.yml up -d
+# go to after all services are started http://localhost
```
bossenti, line 98:
```suggestion
# go to `http://localhost` after all services are started 
```

installer/compose/quickstart/upload.sh
```diff
@@ -0,0 +1,91 @@
...
+        -H "Content-Type: application/json" \
+        -H "Authorization: $RESPONSE_TOKEN")
+    ARRAY_LENGTH=$(echo $JSON_RESOURCE_ITEMS | jq '. | length')
+done
+
+ITEM_MAP_NAME_URI=$(echo $JSON_RESOURCE_ITEMS | jq -r 'map({(.name): .uri}) | add')
+
+ITEM_MAP_NAME_URI=$(echo $JSON_RESOURCE_ITEMS | jq -r 'map({(.name): .uri}) | add')
+
+for ZIP_FILE in /zip_folder/*.zip; do
```
bossenti, line 55: Why do we have multiple ZIP files? Can't we combine them into a single one?

If not, can we please use speaking names for them instead of the UUIDs?

bossenti (commented):
I've had a look at your provided StreamPipes assets in `zip folder` as well. Can you please give both the StreamPipes assets definition and their contained assets themselves more speaking names than `Default-*`. I think this can cause a lot of confusion for first-time users.
In addition, can you please clean up the assets? In one asset group, there is for example an empty dashboard.

installer/compose/README.md
Cccccrystall: Thanks for the suggestion, here actually I am trying to make the format is same as the "full" and "nats" option, I could delete the duplicate row. Should I delete the duplicate "full" and "nats" row together?

bossenti: Yes I think you can delete this row there as well
Thanks for pointing out!

installer/compose/docker-compose.quickstart.yml
Cccccrystall: Hi, @RobertIndie, I have addressed the comment and changed the quickstart based on the docker-compose.yml.

installer/compose/upload.sh
Cccccrystall: Thanks for the suggestion! I have removed the debug codes.

installer/compose/quickstart/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+# limitations under the License.
+# Use an official base image
+FROM ubuntu:latest
+
+# Install curl and jq
+RUN apt-get update \
+    && apt-get install -y curl jq
+
+
+RUN curl -L -o /usr/local/bin/wait-for-it.sh https://raw.githubusercontent.com/vishnubob/wait-for-it/81b1373f17855a4dc21156cfe1694c31d7d1792e/wait-for-it.sh \
```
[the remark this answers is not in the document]
Cccccrystall: Hi, @bossenti ! I have resolved all the commonts by doing the following changes:

1. Removing the logic of using wait-for-it.sh, and implementing the same logic in upload.sh to wait the backend service loading.
2. Enriching the example assets, making the asset and component to contain specific name and description,  cleaning up empty dashboards and redundant elements, offering meaningful dashboards and data view for specific   application scenarios.

installer/compose/README.md
Cccccrystall: Hi, @bossenti ! 
I have refactored the README file, deleting the duplicate rows.

installer/compose/README.md
Cccccrystall: Hi, @bossenti !
I have changed the descriptiong of this row, as well as the same description for 'Nats' and 'Full' option.

installer/compose/quickstart/upload.sh
Cccccrystall: Hi, @bossenti !
I combined the assets into a single ZIP file, and changed the file name.

installer/compose/docker-compose.quickstart.yml
```diff
@@ -0,0 +1,195 @@
...
+    depends_on:
+      - consul
+    volumes:
+      - files:/spImages
+    logging: *default-logging
+    restart: unless-stopped
+    networks:
+      spnet:
+
+  sources-watertank-simulator:
```
bossenti, line 157: We can remove the `sources-watertank-simulator` section here. We will remove them in future releases anyways

installer/compose/quickstart/upload.sh
```diff
@@ -0,0 +1,100 @@
...
+
+if [ -z "$SP_INITIAL_ADMIN_PASSWORD" ]; then
+  PASSWORD="admin"
+else
+  PASSWORD="$SP_INITIAL_ADMIN_PASSWORD"
+fi
+
+while true; do
+  if curl -s "http://backend:8030/streampipes-backend/api/v2/auth/login" --max-time 3 &> /dev/null; then
+    break
```
bossenti, line 32: can we echo something in this case?
In addition, `break` will only stop the loop right? But in this case, we might want to stop the execution of the whole script. If we do so, we can increase `--max-time` as well to 10 or so

bossenti (commented):
@Cccccrystall thanks a lot for addressing the review comments.
Your assets really look good and I think they are greatly suited for the quickstart mode.
However, I have two small remarks with respect to the assets.
* Can we give the file `MOCK_DATA.json` an appropriate name?
* Can we remove the `02` from `Flowrate02` e.g. in `Flowrate02 sensor mass flow condition`

Sorry for being so nit-picking here but I think for the purpose of this improvement such details are important

installer/compose/quickstart/upload.sh
Cccccrystall: Hi @bossenti , thank you for your feedback! Certainly, we can implement a confirmation mechanism within this script. Once the script verifies that the StreamPipes backend is ready and the login API is accessible, it can exit the loop and proceed to the next step. At this point, we can display a message saying, "StreamPipes backend is now ready." Does this approach sound suitable to you?

Regarding exiting the loop, you're correct that using `break` will only terminate the loop, allowing the script to continue with subsequent steps. This is indeed our intention, especially after confirming the readiness of the Streampipes backend under the specified condition. So I think it's better to keep `break` and not stop the whole script.

The `--max-time` option in the curl command is designated to limit the execution time, in seconds, encompassing the duration required to resolve the host, establish a connection, and transfer data. It serves as a precaution against indefinite or excessively long waits due to network complications or an unresponsive target host. Setting this parameter to 10 seconds is a feasible adjustment. I will change it to 10.

bossenti: Ah sorry, I put my comment at the wrong statement...

My only concern was: If we only wait 9s (3 seconds with max-time 3) the backend might not yet be available in some cases. In this case we shouldn't continue with the script  because this would probably lead to errors later (for sure if the backend doesn't get ready in time).
So me asking for termination was only thinking about this special edge case where the backend does not get ready within the specified waiting period.
But when we extend the waiting period to 30 seconds (3 seconds with max-time 10) this should not occur

Cccccrystall: Hi, @bossenti ! Thanks for the suggestion, I will change the `max-time` to 10 in next commit.

installer/compose/docker-compose.quickstart.yml
Cccccrystall: Hi, @bossenti ! 
I have changed the mock data name to `Factory_Environment_Data.json`, removed the `02` and removed `sources-watertank-simulator` section in yml file.

installer/compose/quickstart/upload.sh
Cccccrystall: Hi, @bossenti ! I have changed the `max-time` in latest commit.

bossenti:
Thanks for your updates!

Just recognized that there are some charts in the data explorer where the caption is wrong. E.g. in the image below this is a histogram and not a time series
image

Can you please go through your data views and adjust them accordingly?

bossenti:
Great, everything looks fine now 🙂

Cccccrystall:
Hi, @bossenti ! In the `run-cypress-e2e-tests`, I find that UI test for `Test live data download dialog`  didn't pass, The error shows: `CypressError: `cy.click()` can only be called on a single element. Your subject contained 2 elements. Pass `{ multiple: true }` if you want to serially click each element.` But actually I didn't change the UI file, so I am confused about this error message. Could you give me some suggestions about this?

bossenti:
@Cccccrystall great that you recognized it!
This is not related by any change you have made.
We have some issues with flaky E2E tests in the last time, so no need to worry 🙂

tools/maven/installer.xml
```diff
@@ -30,6 +30,7 @@
             <useDefaultExcludes>true</useDefaultExcludes>
             <excludes>
                 <exclude>%regex[(?!((?!${project.build.directory}/)[^/]+/)*src/)(.*/)?\.gitignore]</exclude>
+                <exclude>/compose/quickstart/zip_folder/*.zip</exclude>
```
bossenti, line 33: `${project.basedir}/compose...`

Cccccrystall: Hi, @bossenti! Thanks for the suggestion! I have updated the path in my latest commit :)

Cccccrystall:
Hi, @bossenti! The recent commit appears to have successfully passed the validation checks, and I am keen to know if there are any parts of improvement you might suggest. Should there be any, I am more than willing to make any necessary revisions promptly to expedite the approval process. 
Furthermore, I want to extend my heartfelt gratitude for taking the time to review my commits and providing suggestions, despite your busy schedule. Your feedback helps in enhancing both my code and documentation. A sincere thank you for your help! ✨

dominikriemer (approved):

Issue #1960, opened 2023-09-24
Better support modification of existing adapters


dominikriemer:
### Body

Currently, it is quite hard to modify existing adapters:
* Adapters can only be edited if no pipeline that uses the data stream is available
* There are several bugs when opening the adapter modification view (e.g., timestamps are duplicated)
* For users, it is quite time-consuming (and frustrating) to delete and re-create all existing pipelines even if the modified adapters doesn't remove existing fields.

To allow convenient modification of adapters, we need to migrate existing resources that depend on an already existing adapter:
* Pipelines need to be re-validated and updated
* Data Lake Measures, used in dashboards and data views, need to be re-validated and updated
* Adapters need to be checked for obsolete transformation rules after modification

In case an adapter is modified which doesn't affect existing resources, all resources should be auto-migrated.
In case an adapter is modified and a dependent resource will break (e.g., a pipeline uses a field that is removed in the adapter), the resource should be flagged for manual investigation/migration.

### StreamPipes Committer

I acknowledge that I am a maintainer/committer of the Apache StreamPipes project.

PR #1979, opened 2023-09-29
feat(#1960): Allow modification of adapters with pipelines

dominikriemer:
### Purpose

* Allow to modify adapters also when they are used by existing pipelines.
* Add auto-migration of pipelines where possible
* Flag invalid pipelines after adapter updates

### Remarks


streampipes-pipeline-management/src/main/java/org/apache/streampipes/manager/recommender/ElementRecommender.java
```diff
@@ -65,11 +65,10 @@ public PipelineElementRecommendationMessage findRecommendedElements() throws NoS
       Optional<SpDataStream> outputStream = getOutputStream(elementsProvider);
       outputStream.ifPresent(spDataStream -> validate(spDataStream, getAll()));
     } catch (Exception e) {
```
bossenti, line 67: Do we want to log the exception at least?

tenthe (approved):

PR #1994, opened 2023-10-03
refactor(#1960): Refactor transformation rules, add schema restore

dominikriemer:
### Purpose

Refactor adapter schema transformation logic: 
* Improve code structure
* Cleanup transformation logic
* Add capability to derive the original schema from transformation rules as a prerequisite for adapter modification (#1960)
* Fix an issue that occurred when adding a new nested field in the schema editor (no runtime name assigned to added field)

### Remarks


streampipes-connect-shared/src/main/java/org/apache/streampipes/connect/shared/preprocessing/generator/StatefulTransformationRuleGeneratorVisitor.java
```diff
@@ -0,0 +1,99 @@
...
+import org.apache.streampipes.model.connect.rules.value.ChangeDatatypeTransformationRuleDescription;
+import org.apache.streampipes.model.connect.rules.value.CorrectionValueTransformationRuleDescription;
+import org.apache.streampipes.model.connect.rules.value.TimestampTranfsformationRuleDescription;
+import org.apache.streampipes.model.connect.rules.value.UnitTransformRuleDescription;
+
+public class StatefulTransformationRuleGeneratorVisitor extends TransformationRuleGeneratorVisitor {
+
+  @Override
+  public void visit(CreateNestedRuleDescription rule) {
+
```
tenthe, line 40: Is there a reason why all the visit methods are emtpy?

streampipes-connect-shared/src/main/java/org/apache/streampipes/connect/shared/preprocessing/utils/ConversionUtils.java
tenthe, line 1: Can we make the methods in this class none static? This is currently not critical, but if we change this, the clients that use the class should be easier to test. This comment also applies to the other Helper and Utils classes. We can also have a discussion about this.

streampipes-model/src/main/java/org/apache/streampipes/model/connect/rules/value/AddTimestampRuleDescription.java
```diff
@@ -42,4 +44,14 @@ public String getRuntimeKey() {
...
   }
+
+  @Override
+  public void accept(ITransformationRuleVisitor visitor) {
+    visitor.visit(this);
+  }
+
+  @Override
+  public int getRulePriority() {
+    return 100;
```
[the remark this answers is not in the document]
tenthe: If this is not possible, I would suggest that we add a small explanation somewhere in the JavaDoc (e.g. at the interface) that briefly explains the concept of rulePriority.

tenthe (approved):

streampipes-model/src/main/java/org/apache/streampipes/model/connect/rules/value/AddTimestampRuleDescription.java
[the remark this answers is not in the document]
dominikriemer: thanks, I modified the code to provide all priorities from an enum

streampipes-connect-shared/src/main/java/org/apache/streampipes/connect/shared/preprocessing/generator/StatefulTransformationRuleGeneratorVisitor.java
dominikriemer: I added a comment to the empty methods to clearly state that the operation is skipped for stateless rules

bossenti (approved):

PR #2024, opened 2023-10-13
fix: make data retrieval of IOLink sensor more robust

bossenti:
### Purpose

This PR makes the data retrieval within the IOLink adapter more robust. Prior to it, an error occurring for one port stopped the processing of all subsequent ports. This means, that in case the processing fails for port 2, e.g., since this port is not connected to a sensor and therefore does not deliver any data, parsing for port 3 and 4 is skipped. 
From now on, only the affected port will be skipped.

### Remarks

### DIFF
--- a/streampipes-extensions/streampipes-connect-adapters-iiot/src/main/java/org/apache/streampipes/connect/iiot/adapters/iolink/IfmAlMqttAdapter.java
+++ b/streampipes-extensions/streampipes-connect-adapters-iiot/src/main/java/org/apache/streampipes/connect/iiot/adapters/iolink/IfmAlMqttAdapter.java
@@ -110,18 +110,26 @@ public class IfmAlMqttAdapter implements StreamPipesAdapter {
 
               for (int i = 0; i < ports.size(); i++) {
 
-                var portResult = getMap(payload,
-                    "/iolinkmaster/port[%s]/iolinkdevice/pdin".formatted(ports.get(i)));
-                var eventData = (String) portResult.get("data");
+                try {
+                  var portResult = getMap(payload,
+                          "/iolinkmaster/port[%s]/iolinkdevice/pdin".formatted(ports.get(i)));
+                  var eventData = (String) portResult.get("data");
 
-                var parsedEvent = sensor.parseEvent(eventData);
-                parsedEvent.put("timestamp", System.currentTimeMillis() + i);
-                parsedEvent.put("port", "port" + ports.get(i));
-                parsedEvent.put(SensorVVB001.IO_LINK_MASTER_SN, serialnumber);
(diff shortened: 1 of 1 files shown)


streampipes-extensions/streampipes-connect-adapters-iiot/src/main/java/org/apache/streampipes/connect/iiot/adapters/iolink/IfmAlMqttAdapter.java
```diff
@@ -110,18 +110,26 @@ public void onAdapterStarted(IAdapterParameterExtractor extractor,
...
-
-                var parsedEvent = sensor.parseEvent(eventData);
-                parsedEvent.put("timestamp", System.currentTimeMillis() + i);
-                parsedEvent.put("port", "port" + ports.get(i));
-                parsedEvent.put(SensorVVB001.IO_LINK_MASTER_SN, serialnumber);
-
-                collector.collect(parsedEvent);
+                try {
+                  var portResult = getMap(payload,
+                          "/iolinkmaster/port[%s]/iolinkdevice/pdin".formatted(ports.get(i)));
```
dominikriemer, line 115: Can we check if the port exists here before fetching it?

streampipes-extensions/streampipes-connect-adapters-iiot/src/main/java/org/apache/streampipes/connect/iiot/adapters/iolink/IfmAlMqttAdapter.java
```diff
@@ -110,18 +110,26 @@ public void onAdapterStarted(IAdapterParameterExtractor extractor,
...
+                  var eventData = (String) portResult.get("data");
```
dominikriemer, line 116: same here

dominikriemer (changes requested):

streampipes-extensions/streampipes-connect-adapters-iiot/src/main/java/org/apache/streampipes/connect/iiot/adapters/iolink/IfmAlMqttAdapter.java
bossenti: Done

streampipes-extensions/streampipes-connect-adapters-iiot/src/main/java/org/apache/streampipes/connect/iiot/adapters/iolink/IfmAlMqttAdapter.java
bossenti: Done

dominikriemer (approved):

Issue #1902, opened 2023-09-09
Implement unit tests for pipeline element 'TextFilterProcessor'


bossenti:
This issue is about implementing unit tests for our pipeline element [TextFilterProcessor](https://github.com/apache/streampipes/blob/752289c6acfc8fa10398c920d678c5c02143ed5c/streampipes-extensions/streampipes-processors-filters-jvm/src/main/java/org/apache/streampipes/processors/filters/jvm/processor/textfilter).
An example of how to write unit tests for pipeline elements can be found [here](https://github.com/apache/streampipes/tree/ca4e0c2d44c1a713caaee3e9c1bff46096f41c53/streampipes-extensions/streampipes-processors-transformation-jvm/src/test/java/org/apache/streampipes/processors/transformation/jvm/processor).

### Mentoring
As this issue is marked as a good first issue and for the `hacktoberfest` event: one of @dominikriemer, @tenthe, @RobertIndie, or @bossenti are happy to provide help for getting started, just tag (one of) them if you want to start working on this issue and need some help.

Aditi840:
@bossenti, can I work on this issue??

Aditi840:
@bossenti, can I work on this issue??

bossenti:
Sure, feel free to go ahead 🙂 

In case you need any assistance we are happy to help!

Aditi840:
@bossenti, will you please give me an overview on what needs to be done.

bossenti:
What are you missing or is unclear to you @Aditi840?
In my opinion, the problem description and that of the parent problem (https://github.com/apache/streampipes/issues/1884) are pretty self-explanatory.

Aditi840:
OK @bossenti, will start working on it.

rahulbiswas876:
@bossenti may I take up the task since it has been more than a month

bossenti:
@rahulbiswas876 you can take on this issue
There is no one assigned currently

rahulbiswas876:
@bossenti  please assign it to me.

PR #2025, opened 2023-10-14
[1902] Implement unit tests for pipeline element 'TextFilterProcessor' 

rahulbiswas876:
### Purpose

- implementation of unit test case for TextFilterProcessor.java class
- see [issue-1902](https://github.com/apache/streampipes/issues/1902)

### Remarks


streampipes-extensions/streampipes-processors-filters-jvm/src/test/java/org/apache/streampipes/processors/filters/jvm/processor/textfilter/TestTextFilterProcessor.java
bossenti, line 1: Does not affect this contribution, but just in case you are eager to contribute further (😉 ):
We now have some boilerplate code that is re-used between multiple unit tests, e.g., `sendEvents`. So it could be worth thinking about some abstraction and generalization here.

bossenti (approved):

PR #2066, opened 2023-10-20
refactor: remove legacy demo resources

bossenti:
### Purpose

This PR removes some legacy resources that have been used for demo purposes but are rather outdated.

### Remarks

PR introduces (a) breaking change(s): yes (see below)

### Breaking Change

This PR removes two data sources from StreamPipes:

* `watertank-simulator`
* `vehicle-simulator`


bossenti:
follow-up task: https://github.com/apache/streampipes/issues/2067

dominikriemer:
@bossenti if I'm not mistaken the PR not only removes the demo sources, but also the `streampipes-sources` module. This includes methods to define data streams programmatically. I think this is a useful feature for use cases where streams should be created directly instead of defining adapters.
Adapters are useful for configurable and reusable connectors, but there might be cases where a fixed data stream should be defined using the SDK. 
Do we really want to delete this?

bossenti:
@dominikriemer Good point!

I deleted `streampipes-sources` because it only contains two classes:

* `AbstractAdapterIncludedStream`
* `AbstractAlreadyExistingStream`

Both only have been used by the two demo sources removed within this PR so I considered them to be legacy 😅 So are they still required to create a data stream? If my understanding is correct, a StreamPipes Function uses the the client to generate a new data stream, for example.

dominikriemer:
Yes, I think they are not directly needed by the function but users can use these convenience classes to create data streams in the SDK. These are the classes we currently mention in the data stream tutorial (https://streampipes.apache.org/docs/extend-tutorial-data-sources/).
I'd suggest to keep these classes and just remove the demo resources.

bossenti:
Okay, then let's keep them.
Should we leave them in `streampipes-sources` or move them elsewhere?

dominikriemer:
@bossenti can it be moved to streampipes-extensions-management? Since they are helper classes similar to the PullAdapter.

dominikriemer (approved):

tenthe (approved):

Issue #2001, opened 2023-10-05
Support migration of adapter and pipeline element configurations


dominikriemer:
### Body

Currently, no mechanism exists to update the configuration of pipeline elements. For instance, when a new configuration is added to an adapter or data processor, existing processors can stop working since the configuration of installed pipeline elements and pipelines which use these elements is broken.

I will open a discussion for the approach to migrate pipeline elements, this issue collects the subtasks to implement such a concept:

- [x] #2002
- [x] Develop migration concept
- [x] #2045
- [x] #2104

### StreamPipes Committer

I acknowledge that I am a maintainer/committer of the Apache StreamPipes project.

bossenti:
## Introduce the ability to migrate existing adapters & PE

>[!IMPORTANT]
> The current implementation only allows modifications that affect the configuration of 
> an element. Changes that affect the event stream are out of scope for this iteration
> (see more at next steps).

This can be split into major parts:

- update description for new elements: adapter, processor, sinks
- update descriptions for existing elements: adapter, processor, sinks

### Rationale
As of now, a user cannot simply upgrade to a recent StreamPipes version if this version introduces changes in a pipeline element (adapter, processor, or sink). To create new instances of an affected pipeline element the user needs to update them first via the UI.
Secondly, existing instances of an affected pipeline element were required to be recreated manually.
Therefore, we now introduce so-called migrations to update both the descriptions used for new elements and the existing elements automatically in the background.

### General Approach

A developer is now enabled to define one or multiple `Migration` for a pipeline element.
Therefore, one of `AdapterMigrator`, `DataProcessorMigrator`, or `DataSinkMigrator` from `streampipes-extensions-api` needs to be implemented. In addition, the migration needs to be added to the service definition in the extensions model submitter, e.g., `AllExtensionsIIoTInit`.

From now on, the migration is fully automatically handled according to the following flow:

![migration-flow](https://github.com/apache/streampipes/assets/50115603/2cf1cfb6-f6fe-4ae9-9b3f-ebd0bfa9b656)

### Update Description for New Elements

For every migration that is sent from the extensions service to the core, the core does update the description of the corresponding element in the storage (couchDB). This is done by requesting the current configuration from the extensions service and overriding the existing one in the storage. This automatic update of all element descriptions makes the manual update in the UI redundant. Thus we can remove this functionality as an upcoming step.

### Update Descriptions for Existing Elements

[...]

PR #2077, opened 2023-10-23
implement migration for adapter and pipeline element configurations

bossenti:
### Purpose

see #2045 

See [detailed explanation](https://github.com/apache/streampipes/issues/2001#issuecomment-1782520174)

### Remarks

### Deprecations

- Deprecate `create()` methods of `ProcessingElementBuilder` that do not contain a version
- Deprecate `create()` methods of `AdapterConfigurationBuilder` that do not contain a version


streampipes-extensions-api/src/main/java/org/apache/streampipes/extensions/api/migration/AdapterMigrator.java
```diff
@@ -0,0 +1,33 @@
...
+import org.apache.streampipes.model.connect.adapter.AdapterDescription;
+
+public abstract class AdapterMigrator implements ModelMigrator<AdapterDescription, IStaticPropertyExtractor> {
+
+  @Override
+  public boolean equals(Object obj) {
+    if (obj instanceof ModelMigrator<?, ?>) {
+      return this.config().equals(((ModelMigrator<?, ?>) obj).config());
+    }
+    return false;
```
dominikriemer, line 31: Just out of curiosity, where is this `equals` method used? Is this the only reason the `AdapterMigrator` has to be an abstract class instead of an interface?

streampipes-client/src/main/java/org/apache/streampipes/client/api/AdminApi.java
```diff
@@ -98,4 +112,16 @@ private StreamPipesApiPath getFunctionsPath() {
   private StreamPipesApiPath getDeleteFunctionPath(String functionId) {
     return getFunctionsPath().addToPath(functionId);
   }
+
+  private StreamPipesApiPath getAdapterMigrationPath() {
+    return StreamPipesApiPath
+            .fromBaseApiPath()
+            .addToPath("migrations/adapter");
```
dominikriemer, line 119: I suggest to use `addToPath` for every URL part: `addToPath("migrations").addToPath("adapter") and then to reuse the migrations path in a constant

streampipes-model/src/main/java/org/apache/streampipes/model/graph/DataProcessorInvocation.java
```diff
@@ -67,28 +69,31 @@ public DataProcessorInvocation(DataProcessorInvocation other) {
     }
     this.pathName = other.getPathName();
     this.category = new Cloner().epaTypes(other.getCategory());
+    this.serviceTagPrefix = SpServiceTagPrefix.DATA_PROCESSOR;
   }
 
   public DataProcessorInvocation(DataProcessorDescription sepa, String domId) {
     this(sepa);
     this.dom = domId;
+    this.serviceTagPrefix = SpServiceTagPrefix.DATA_PROCESSOR;
```
dominikriemer, line 78: Can this line be removed since it's already assigned in the called constructor?

streampipes-model/src/main/java/org/apache/streampipes/model/graph/DataSinkInvocation.java
```diff
@@ -51,16 +53,19 @@ public DataSinkInvocation(DataSinkDescription other) {
     this.setIncludesAssets(other.isIncludesAssets());
     this.setElementId(ElementIdGenerator.makeElementId(this));
     this.setIncludedAssets(other.getIncludedAssets());
+    this.serviceTagPrefix = SpServiceTagPrefix.DATA_SINK;
   }
 
   public DataSinkInvocation(DataSinkDescription sec, String domId) {
     this(sec);
     this.setDom(domId);
+    this.serviceTagPrefix = SpServiceTagPrefix.DATA_SINK;
```
dominikriemer, line 62: same here

dominikriemer:
@bossenti Is there a reason why we have two endpoints at the core side for managing migrations of adapters and pipeline elements?
Does it have any implications if we remove the filtering logic from `ExtensionsModelSubmitter` to the core and have a single migration method there for adapters and pipeline elements?
This would ease managing the status of services.

dominikriemer:
I create a PR which breaks down the migration into smaller code units and creates a single request from extensions to core for submitting migration configs: #2101

tenthe (approved):

bossenti:
@dominikriemer you already addressed all your comments in your PR #2101, right?

dominikriemer:
@bossenti I didn't address the comments above - do you know where the equals method is used from the `AdapterMigrator` or can we change it to an interface?

bossenti:
@dominikriemer Yes it is. We use it to determine if a migration has already been registered at the extensions service so that we can't register the same migration multiple times.

Issue #2096, opened 2023-10-27
Assets need to be updated as well (only for adapters)


bossenti:
### Body

Within https://github.com/apache/streampipes/issues/2001 we introduced basic support for migrating pipeline elements.
One aspect that is not yet covered is the update of the corresponding assets (string values, documentation & icon) for adapters.
So we need to include the update of the assets in the migration process of an adapter at the core.

### StreamPipes Committer

I acknowledge that I am a maintainer/committer of the Apache StreamPipes project.

bossenti:
closed by https://github.com/apache/streampipes/pull/2105

PR #2105, opened 2023-10-30
Improve startup behaviour

dominikriemer:
### Purpose

* Improve the lifecycle of various requests during the initialization of core and extensions
* Introduce states for extension services REGISTERED, MIGRATING, HEALTHY, UNHEALTHY
* Introduce states for core INSTALLING, MIGRATING; READY
* Add logic to improve the order of initialization actions: Start core, register extensions, perform core migrations, perform extenion migrations

### Remarks


streampipes-service-extensions/src/main/java/org/apache/streampipes/service/extensions/CoreRequestSubmitter.java
```diff
@@ -0,0 +1,54 @@
...
+
+public class CoreRequestSubmitter {
+
+  private static final Logger LOG = LoggerFactory.getLogger(CoreRequestSubmitter.class);
+
+  private static final int RETRY_INTERVAL_SECONDS = 3;
+
+  public void submitRepeatedRequest(Supplier<Boolean> request,
+                                    Supplier<String> successMessage,
+                                    Supplier<String> failureMessage) {
```
bossenti, line 37: Can't we pass the messages as simple strings?

streampipes-pipeline-management/src/main/java/org/apache/streampipes/manager/migration/AdapterDescriptionMigration093.java
```diff
@@ -0,0 +1,73 @@
...
+          .stream()
+          .filter(tag -> tag.getPrefix() == SpServiceTagPrefix.ADAPTER)
+          .filter(tag -> appIdsToReinstall.contains(tag.getValue()))
+          .forEach(tag -> {
+            var appId = tag.getValue();
+            try {
+              if (adapterDescriptionStorage.getAdaptersByAppId(appId).isEmpty()) {
+                var requestUrl = getRequestUrl(SpServiceTagPrefix.ADAPTER, appId, serviceUrl);
+                var entityPayload = HttpJsonParser.getContentFromUrl(URI.create(requestUrl));
+                Operations.verifyAndAddElement(
```
bossenti, line 62: Does this then also include assets (icons, etc.)? If so, this would close #2096

tenthe (approved):

PR #2109, opened 2023-10-30
feat: extend FileApi with the capability to query by the original file name

bossenti:
### Purpose

This PR extends the FileAPI of the core and the client to query by the original file name.

### Remarks


bossenti:
Would you rather like to see the `IFileAPI` interface untouched and simply extend the file client with an additional method or are you fine with the proposed solution?

tenthe:
Hey @bossenti,
I think it is ok to change the interface. But I would suggest to add a new method. Then it is not a breaking change for clients who already use the interface.
Further, I would suggest to rename the parameter of the existing methods to make clear that an internal id is used instead of the file name.

How can we deal with files that have the same name? Any ideas?

Cheers,
Philipp

dominikriemer:
How do we distinguish between files with the same  original filename and which file will be delivered in this case? I think in this case we would need to ensure that no two files with the same filename can be uploaded.

tenthe:
@dominikriemer yes, I totally agree. I think we should make sure that there are no two files with the same name.
Currently, if a user uploads two files with the same name and wants to delete one of them, it is not clear which one is which.
I think the best solution for this is to only allow unique names.
But I would address this in a separate issue and PR.

bossenti:
I have simplified the API a bit so that we don't have the adapt all usages of `getFileContent`.
Let's take the uniqueness of uploaded files as a follow-up issue.

tenthe (approved):

Issue #2122, opened 2023-11-02
Cleanup extension bundles


dominikriemer:
### Body

Currently, we provide many different bundles of extensions as fat jars and docker images:
* Connect adapters
* Connect adapters IIoT
* Pipeline elements all jvm
* Extensions all jvm
* Extension IIoT jvm
* and more

We can harmonize these to ease the installation.

I suggest the following three "flavors" that we can offer in the standard installation:

* `extensions-all-jvm`: Includes all adapters and pipeline elements in the codebase, as well as support for all protocols (Kafka, JMS, Pulsar, MQTT, Nats)
* `extensions-all-iiot` Includes a subset of pipeline elements and adapters which are relevant for IIoT scenarios, e.g., excluding image processing and geo components and including support for all protocols
* `extensions-iiot-minimal`: A docker image minized for small image size with reduced dependencies, only relevant IIoT adapters and sinks, and only providing support for Nats as a broker.

This would allow us to also ship smaller installations so that first-time users don't need to wait too long when pulling all the docker images.

### StreamPipes Committer

I acknowledge that I am a maintainer/committer of the Apache StreamPipes project.

bossenti:
@dominikriemer I really like the three flavors proposed by you!

PR #2124, opened 2023-11-02
feat(#2122): Reorganize extension modules and modify docker images

dominikriemer:
### Purpose

See #2122 

### Remarks

* Remove pipeline-elements-all-jvm module and Docker image, which was used in the `full` installer


streampipes-extensions/streampipes-connect-adapters/src/main/java/org/apache/streampipes/connect/GeneralAdaptersExtensionModuleExport.java
```diff
@@ -28,35 +28,42 @@
...
+import java.util.List;
 
-  public static void main(String[] args) {
-    new ConnectAdapterInit().init();
+public class GeneralAdaptersExtensionModuleExport implements IExtensionModuleExport {
+
+  @Override
+  public List<StreamPipesAdapter> adapters() {
+    return List.of(
+        //new GdeltAdapter(),
```
tenthe, line 44: I guess we can remove the comments here, right? I would also suggest removing the adapters.

tenthe (approved):

streampipes-extensions/streampipes-connect-adapters/src/main/java/org/apache/streampipes/connect/GeneralAdaptersExtensionModuleExport.java
tenthe: I would suggest that we open an issue to deprecate some of the adapters in the `streampipes-connect-adapters` module for the upcoming release, so that we can remove them in 0.94.0. 
The module has several adapters that are out of date.

dominikriemer: Yes, we can do that in a separate PR

installer/compose/docker-compose.minimal.yml
```diff
@@ -0,0 +1,146 @@
...
+      spnet:
+
+  nats:
+    image: nats
+    restart: unless-stopped
+    networks:
+      spnet:
+
+  # Lite version pipeline elements
+  extensions-all-jvm:
```
bossenti, line 121: Do we need to call the service `extensions-all-jvm`?

streampipes-rest-extensions/src/main/java/org/apache/streampipes/rest/extensions/migration/DataSinkMigrationResource.java
```diff
@@ -47,7 +45,6 @@ public class DataSinkMigrationResource extends MigrateExtensionsResource<
   @POST
   @Consumes(MediaType.APPLICATION_JSON)
   @JacksonSerialized
-  @PreAuthorize(AuthConstants.IS_ADMIN_ROLE)
```
bossenti, line 50: Is this check redundant or why do we want to remove it?

dominikriemer: I removed this for now because this functionality required a dependency to `streampipes-rest` in the  `streampipes-rest-extensions` module, which pulled in all core dependencies in the extension modules, adding a lot of overhead.
Since there's currently no authorization for extensions in place, we can address this later if needed.

installer/compose/docker-compose.minimal.yml
dominikriemer: Good catch, will change that

PR #2132, opened 2023-11-03
improvement: Optimize startup behaviour

dominikriemer:
### Purpose

Add some improvements to startup behaviour for new installations and migrations.

### Remarks


streampipes-config/src/main/java/org/apache/streampipes/config/backend/BackendConfig.java
```diff
@@ -128,8 +126,4 @@ private String getJwtSecret() {
   private String makeDefaultJwtSecret() {
     return TokenGenerator.generateNewToken();
   }
-
-  public void updateSetupStatus(boolean status) {
-    config.setBoolean(BackendConfigKeys.IS_SETUP_RUNNING, status);
```
bossenti, line 133: `BackendConfigKeys.IS_SETUP_RUNNING` is not used anywhere else, right?
Thus we can remove it

bossenti (approved):

Issue #2135, opened 2023-11-03
Max health check intervals configurable


dominikriemer:
### Body

Currently, intervals for health checks are statically defined. The intervals should be configurable by environment variables.

### StreamPipes Committer

I acknowledge that I am a maintainer/committer of the Apache StreamPipes project.

PR #2136, opened 2023-11-03
feat(#2135): Max health check intervals configurable

dominikriemer:
### Purpose

See #2135 

### Remarks

PR introduces (a) breaking change(s):  no


streampipes-service-core/src/main/java/org/apache/streampipes/service/core/StreamPipesCoreApplication.java
```diff
@@ -133,45 +129,54 @@ public void init() {
...
 
   private boolean isConfigured() {
     return new UserStorage().existsDatabase();
   }
 
-  private void doInitialSetup() {
+  private void doInitialSetup(int initialSleepBeforeInstallation) {
     LOG.info("\n\n**********\n\nWelcome to Apache StreamPipes!\n\n**********\n\n");
     LOG.info("We will perform the initial setup, grab some coffee and cross your fingers ;-)...");
     LOG.info("Auto-setup will start in 5 seconds to make sure all services are running...");
```
bossenti, line 177: `5 seconds` needs to be printed dynamically

tenthe (approved):

Issue #1936, opened 2023-09-14
StreamPipes Connect: Ease deletion of adapters


dominikriemer:
### Body

Users should not be able to delete adapters which are used by an existing pipeline.
The current dialog prevents deletion of such adapters, but has some drawbacks:
- [x] There is no hint to the name of the pipeline which uses the adapter, just the number
https://github.com/apache/streampipes/pull/2070
- [x] From a usability point of view, it would be better to ask users if the related pipelines should be deleted as well. In this case, adapters and pipelines can be deleted with a single click without needing to navigate to the pipeline view, find the right pipeline, delete the pipeline, and go back to the adapter view.
https://github.com/apache/streampipes/pull/2070
- [x] It is possible that the user who owns an adapter has shared the adapter with other users and the adapter owner does not have access rights to this pipeline. Pipelines that are created by other users using the adapter should not be deleted, in this case an error message should appear.

### StreamPipes Committer

I acknowledge that I am a maintainer/committer of the Apache StreamPipes project.

muyangye:
Hi @dominikriemer @bossenti may I work on this issue? I have traced through the code a bit and I am confident I can do the first 2 improvements (hint the names and ask users if related pipelines should be deleted as well). I have not looked into the 3rd improvement yet but anyway I think it can only be done after the 2nd improvement is done. Thanks!

bossenti:
Hi @muyangye,

yes, that would be awesome! I agree with you that an iterative approach probably works best here. Please be aware of @dominikriemer PR which is currently still WIP but has some implications for this issue: https://github.com/apache/streampipes/pull/1979. So please have a look there before starting on this issue.

Please feel free to ask for advice in case you don't know how to proceed or you are facing any errors. We are happy to help 🙂

dominikriemer:
Hi @muyangye that would be awesome!
Please ask if you need any help.

muyangye:
Thanks for reminding me the WIP PR, that is indeed very relevant to this issue. One quick question, how can I see my log messages in module `streampipes-rest` and module `streampipes-storage-couchdb`? I used to do "System.out.println()" but this time no message is shown on the consoles of `all-pipeline-elements-jvm` or `connect-adapters` probably because they are not the correct consoles that have my log. I also tried using the log4j Logger but I still don't know where to locate my log message. Can you please tell me where can I find the log for these 2 modules as this information is valuable for me to proceed? Thank you very much!

bossenti:
[...]

PR #2139, opened 2023-11-05
[#1936] Adapter Deletion: Disallow Deletion of Pipelines Using Adapter but Not Owned by User

muyangye:
### Purpose

See #1936 

### Remarks


muyangye:
Currently, even admin can't delete pipelines created by others because I am comparing the id of pipeline's owner with the id of the user. Let me know if admin should delete pipelines not owned by himself.

dominikriemer:
Hi @muyangye great that you're working on this topic!
Yes, I think admin users should always be able to delete resources owned by others.

muyangye:
> Hi @muyangye great that you're working on this topic!
> Yes, I think admin users should always be able to delete resources owned by others.

Can they also delete resources owned by another admin?

tenthe:
Hello @muyangye,
thanks for your work on this topic. 
In my opinion, it makes perfect sense for an admin to have the right to delete all resources, even if they were created by another admin.

muyangye:
@dominikriemer @tenthe, thanks for clarifying! Just pushed an update that addressed this.

tenthe (approved):
@muyangye, thanks a lot for the PR.
The feature looks good

dominikriemer (approved):

Issue #2095, opened 2023-10-27
Remove update functionality from UI


bossenti:
### Body

Within https://github.com/apache/streampipes/issues/2001 we introduced basic support for migrating pipeline elements.
This makes the manual update of pipeline elements in the UI redundant.
Therefore, we can safely remove it.

### StreamPipes Committer

I acknowledge that I am a maintainer/committer of the Apache StreamPipes project.

bossenti:
<img width="508" alt="image" src="https://github.com/apache/streampipes/assets/50115603/a4a29975-aa52-4170-8d68-927e2d386658">

PR #2155, opened 2023-11-08
refactor(#2095): remove update button & move permissions button

bossenti:
### Purpose

see #2095 

One pipeline element is displayed the following now:
image

### Remarks

### DIFF
--- a/ui/src/app/add/components/endpoint-item/endpoint-item.component.html
+++ b/ui/src/app/add/components/endpoint-item/endpoint-item.component.html
@@ -90,34 +90,18 @@
                 </button>
                 <div *ngIf="item.installed" style="margin-left: 5px">
                     <button
-                        class="small-button-add mat-basic no-min-width"
-                        [disabled]="!item.editable"
-                        mat-raised-button
+                        class="small-button-add-inverted"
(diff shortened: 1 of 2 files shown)


dominikriemer:
@bossenti one question, how will development work without the update button?
E.g., when I develop a pipeline element, I often modify configs and test them by using the update button. Will this still work with the new approach or would I need to increase the version with every change?

bossenti:
We do overwrite all adapter & pipeline element descriptions when starting an extensions service right? 
If so, there should be no need to manually update the descriptions during development. Or am I mistaken?

tenthe:
@bossenti I'm not entirely certain if this is already the case, but I agree with you on how it should be.

tenthe (approved):

bossenti:
I'll give it a final check before merging

bossenti:
Which one do we want to take?

dominikriemer:
I'd prefer to keep the manual update button for now.
But I like the `Manage Permissions` button ;-)

bossenti:
Okay, fine for me 
Me too, but there is no space for three buttons right?

Issue #2166, opened 2023-11-10
CSV reader in FileStream adapter cannot handle strings


bossenti:
### Apache StreamPipes version

dev (current development state)

### Affected StreamPipes components

Connect

### What happened?

When reading the following CSV file with the `FileStream` adapter and persisting the created the data stream in the data storage, simple string values are lost.
Only string values that are explicitly quoted are kept, but this should not be necessary for a CSV file.

```text
ts,testId,name,assetId
1000,1,Tom,“5“
2000,2,Albert,“1“
```

### How to reproduce?

<img width="1636" alt="image" src="https://github.com/apache/streampipes/assets/50115603/da9b9585-027d-48b1-92e8-7ef11ce1d18b">

<img width="750" alt="image" src="https://github.com/apache/streampipes/assets/50115603/b19dbdeb-3103-452c-beac-4cddbd62f842">

<img width="948" alt="image" src="https://github.com/apache/streampipes/assets/50115603/39abcae0-2929-44b3-92b8-212325e4306a">

### Are you willing to submit a PR?

None

muyangye:
Hi @bossenti, I have looked into the issue. The event is "correct" ("correct" meaning key value pairs are the same with the file) in frontend, Kafka producer, and Kafka consumer. However, once the code hits [Influx Store](https://github.com/apache/streampipes/blob/8bb9168c0e6e7e1770e1b9d1291978d75bd00890/streampipes-data-explorer-commons/src/main/java/org/apache/streampipes/dataexplorer/commons/influx/InfluxStore.java), the runtime name of some keys are modified. The root cause is that "name" key in the file stream is a keyword of [Influx reserved keywords list](https://github.com/apache/streampipes/blob/8bb9168c0e6e7e1770e1b9d1291978d75bd00890/streampipes-data-explorer-commons/src/main/java/org/apache/streampipes/dataexplorer/commons/influx/InfluxDbReservedKeywords.java) and Streampipes [sanitizes](https://github.com/apache/streampipes/blob/8bb9168c0e6e7e1770e1b9d1291978d75bd00890/streampipes-data-explorer-commons/src/main/java/org/apache/streampipes/dataexplorer/commons/influx/InfluxNameSanitizer.java#L21) DataLake measure for those keys conflicting with Influx reserved keywords. If you change "name" to something else not in the list everything works.

To fix this, I propose to also sanitize event instead of only sanitizing DataLake measure. Let me know what do you think🙂!

[...]

PR #2172, opened 2023-11-13
fix(#2166): sanitize event keys in influx sink

muyangye:
### Purpose

See #2166 

### Remarks


bossenti (approved):
Great improvement @muyangye 🙏🏼

bossenti:
Works like a charm!

Issue #690, opened 2022-11-26
Remove lodash dependency from ui


dominikriemer:
The lodash dependency is used in "PipelineElementOptionsComponent" where the cloneDeep function is imported.

 

 

Imported from Jira [STREAMPIPES-525](https://issues.apache.org/jira/browse/STREAMPIPES-525). Original Jira may contain additional context.
Reported by: zehnder.

mohitbhagat:
Hi @bossenti @dominikriemer , I'm not sure what's expected here. I notice some usages of lodash in `pipeline-element-options.component.ts`. We are using cloneDeep from lodash, to do some deep cloning. Am I expected to replace cloneDeep with a custom clone method? What's the reason behind wanting to get rid of the lodash dependency?

bossenti:
Unfortunately, I cannot provide any support here.
@dominikriemer probably knows best about this topic

PR #2193, opened 2023-11-16
Remove lodash dependency from UI

Marcelfrueh:
### Purpose
 Remove lodash from UI by replacing cloneDeep function with newly added deepCopy function. Moreover lodash is removed from the dependency list.

### Remarks
 https://github.com/apache/streampipes/issues/690


ui/src/app/editor/components/pipeline-element-options/pipeline-element-options.component.ts
```diff
@@ -227,4 +226,25 @@ export class PipelineElementOptionsComponent implements OnInit, OnDestroy {
...
     }
+
+    deepCopy(obj) {
+        const clone = {};
+
+        if (
+            typeof obj !== 'object' ||
+            typeof obj === undefined ||
+            obj === null ||
+            Array.isArray(obj) ||
```
dominikriemer, line 237: Just a question...why are arrays excluded here?
E.g., when called with a `PipelineElementConfig[]` object (line 164), it seems the method returns the original input and not a deep copy of the list including its elements?

dominikriemer (approved):

tenthe (approved):

Issue #2106, opened 2023-10-30
Influx Sink can only handle primitive properties


bossenti:
### Apache StreamPipes version

dev (current development state)

### Affected StreamPipes components

Processing Elements

### What happened?

When trying to persist an event stream with an array as property, the influx sink throws an exception:

> 2023-10-30T13:25:28.623+01:00 ERROR 25805 --- [       Thread-3] >o.a.s.w.s.r.StandaloneEventSinkRuntime   : RuntimeException while processing event in >org.apache.streampipes.sinks.internal.jvm.datalake.DataLakeSink
>
>java.lang.IllegalArgumentException: Expecting a positive number for fields size
>	at org.influxdb.impl.Preconditions.checkPositiveNumber(Preconditions.java:35)
>	at org.influxdb.dto.Point$Builder.build(Point.java:352)
>	at org.apache.streampipes.dataexplorer.commons.influx.InfluxStore.onEvent(InfluxStore.java:176)
	at org.apache.streampipes.dataexplorer.commons.TimeSeriesStore.onEvent(TimeSeriesStore.java:64)
	at org.apache.streampipes.sinks.internal.jvm.datalake.DataLakeSink.onEvent(DataLakeSink.java:82)
	at org.apache.streampipes.wrapper.standalone.runtime.StandaloneEventSinkRuntime.process(StandaloneEventSinkRuntime.java:55)
	at org.apache.streampipes.wrapper.standalone.routing.StandaloneSpInputCollector.send(StandaloneSpInputCollector.java:56)
	at org.apache.streampipes.wrapper.standalone.routing.StandaloneSpInputCollector.lambda$onEvent$0(StandaloneSpInputCollector.java:51)
	at java.base/java.util.concurrent.ConcurrentHashMap.forEach(ConcurrentHashMap.java:1603)
	at org.apache.streampipes.wrapper.standalone.routing.StandaloneSpInputCollector.onEvent(StandaloneSpInputCollector.java:51)
	at org.apache.streampipes.wrapper.standalone.routing.StandaloneSpInputCollector.onEvent(StandaloneSpInputCollector.java:30)
	at org.apache.streampipes.messaging.kafka.SpKafkaConsumer.lambda$run$0(SpKafkaConsumer.java:105)
	at java.base/java.lang.Iterable.forEach(Iterable.java:75)
	at org.apache.streampipes.messaging.kafka.SpKafkaConsumer.run(SpKafkaConsumer.java:105)
[...]

PR #2196, opened 2023-11-17
fix(#2106) Serialize non-primitive types as string and store in InfluxDB

muyangye:
### Purpose

See #2106 

### Remarks

I treated ListField and NestedField as the same and used Jackson to serialize them into string.

For this test adapter: 9a64bd488ff1a0cc8cd422783dc4e13
Now we can see the raw value being stored (but not other visualizations such as table because they are not implemented yet): 
cf10c485fdb20dbc6ea6285610d1ddd
The raw value doesn't show deserialization result, so I have left some commented out codes that do the deserialization and below shows that the original types are kept:
808b41dab8710524cff2b57aca2f930

### **HOWEVER**
I am a little skeptical about my approach so please let me know if this is okay. I am somewhat confused on the difference between serializing Abstract/Primitive/List/Nested Field vs. serializing raw values. The reasons I chose to serialize raw values are:

- Serializing Abstract/Primitive/List/Nested Field which includes inheritance is very hard, I have found multiple resources online but no luck to get the correct result.
- Later I found out the serialization/deserialization of those Fields are unnecessary because InfluxDB is actually storing runtime name and raw value.

Let me know if it is necessary to serialize those Fields. Currently I don't think so because during deserialization, we can check `instanceof List` and `instanceof Map` for `Object`, just like we use `instanceof ListField` and `instanceof NestedField` for `AbstractField`.


streampipes-data-explorer-commons/src/main/java/org/apache/streampipes/dataexplorer/commons/influx/InfluxStore.java
```diff
@@ -218,6 +226,21 @@ private void handleMeasurementProperty(Point.Builder p,
     }
   }
 
+  private void handleNonPrimitiveMeasurementProperty(Point.Builder p, Event event, String preparedRuntimeName) {
+    var field = event.getRaw().get(preparedRuntimeName);
+    ObjectMapper mapper = new ObjectMapper();
```
tenthe, line 231: I think we can change the object mapper as a class variable.

streampipes-data-explorer-commons/src/main/java/org/apache/streampipes/dataexplorer/commons/influx/InfluxStore.java
```diff
@@ -218,6 +226,21 @@ private void handleMeasurementProperty(Point.Builder p,
...
+    try {
+      String json = mapper.writeValueAsString(field);
+      p.addField(preparedRuntimeName, json);
+      // Deserialize in this way, tested the primitive types can be correctly deducted from Object
+//      TypeFactory typeFactory = mapper.getTypeFactory();
```
tenthe, line 236: If the code is working, I would remove this comment before we merge it. Is there a way we can unit test it? I think in the current version of this class this will be hard. But maybe we can open another issue to implement unit tests for this class and then refactor it. What do you think?

streampipes-data-explorer-commons/src/main/java/org/apache/streampipes/dataexplorer/commons/influx/InfluxStore.java
```diff
@@ -218,6 +226,21 @@ private void handleMeasurementProperty(Point.Builder p,
     }
   }
 
+  private void handleNonPrimitiveMeasurementProperty(Point.Builder p, Event event, String preparedRuntimeName) {
+    var field = event.getRaw().get(preparedRuntimeName);
+    ObjectMapper mapper = new ObjectMapper();
+    try {
+      String json = mapper.writeValueAsString(field);
```
tenthe, line 233: One more generic comment. We started to switch to the new `var` syntax. Could you please do the same when you make changes to the code. I know we have not yet implemented this consistently everywhere, but we want to adapt it bit by bit.

tenthe (changes requested):
Thank you very much for providing the PR. That looks very good.
And I think this solution is sufficient for the moment.
I have some minor comments we can discuss below.
Cheers,
Philipp

streampipes-data-explorer-commons/src/main/java/org/apache/streampipes/dataexplorer/commons/influx/InfluxStore.java
muyangye: Will do.

streampipes-data-explorer-commons/src/main/java/org/apache/streampipes/dataexplorer/commons/influx/InfluxStore.java
muyangye: I agree, probably we should refactor the serialization/deserialization logic into a class called `FieldSerializer`, in which the `Serialize` method takes an `AbstractField` and returns a `String` and the `Deserialize` method takes a `String` and return a `Map<string, deserialized type (as opposed to Object in my comments)>`. For deserialization, I plan to write a recursive function in which the base case is when the value is primitive or List. Do you want me to write all that logic in this PR?

muyangye:
Hi @tenthe, I have pushed an update according to your review. Specifically, I made a class `RawFieldSerializer` that takes care of both serialization and deserialization. I have also added unit tests for that class. In addition, I have scanned through `InfluxStore` and changed all variable declarations to `var`.

tenthe (approved):
@muyangye thanks a lot

Issue #2151, opened 2023-11-08
 Implement E2E Tests for Enhanced Adapter Deletion


tenthe:
### Body

Issue #1936 introduced several new features related to adapter deletion. 
To ensure the robustness of these new functionalities, we need to implement End-to-End (E2E) tests that cover all aspects of adapter deletion.

### StreamPipes Committer

I acknowledge that I am a maintainer/committer of the Apache StreamPipes project.

muyangye:
Let me implement these tests since it's a relatively new feature!

PR #2197, opened 2023-11-17
#2151 & #2223: Implement E2E tests for enhanced adapter deletion and Extend reset REST endpoint to remove users

muyangye:
### Purpose

See #2151 

### Remarks

Added E2E tests for these 3 cases:

- An adapter and >1 pipelines are created by a non-admin user, the non-admin user should be able to delete the adapter and associated pipelines.
- A public adapter and >1 pipelines are created by a non-admin user, the admin should be able to delete the adapter and associated pipelines.
- A public adapter and >1 pipelines are created by admin (or any other users), the non-admin user should not be able to delete the adapter and the associated pipelines. After checking this, switch to admin's account to clean up (delete adapter, associated pipelines, non-admin user).

NOTE: I modified the logic of roles in `addUser()` of `UserUtil.ts` because I need the non-admin user to have both Connect and Pipeline admin privileges. I have run cypress tests locally and they all passed but I am not sure if this change will affect other tests in CI. For some reason the original logic is only taking the first role out of all roles.

Also sorry for closing my previous PR because I force pushed it to revert changes and separate my PR to 2 PRs 😅==

### DIFF
--- a/ui/cypress/tests/adapter/enhancedDeleteAdapter.smoke.spec.ts
+++ b/ui/cypress/tests/adapter/enhancedDeleteAdapter.smoke.spec.ts
@@ -48,7 +48,7 @@ const pipelineInput = PipelineBuilder.create('Pipeline Test')
 describe('Test Enhanced Adapter Deletion', () => {
     // In the beginning, create the non-admin user
     before(() => {
-        cy.login(UserUtils.adminUser);
+        cy.initStreamPipesTest();
         UserUtils.addUser(UserUtils.adapterAndPipelineAdminUser);
         cy.logout();
     });
--- a/ui/cypress/tests/adapter/enhancedDeleteAdapter.smoke.spec.ts
+++ b/ui/cypress/tests/adapter/enhancedDeleteAdapter.smoke.spec.ts
@@ -1,0 +1,122 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+import { AdapterBuilder } from '../../support/builder/AdapterBuilder';
+import { ConnectUtils } from '../../support/utils/connect/ConnectUtils';
+import { PipelineBuilder } from '../../support/builder/PipelineBuilder';
+import { PipelineElementBuilder } from '../../support/builder/PipelineElementBuilder';
+import { UserUtils } from '../../support/utils/UserUtils';
+import { PipelineUtils } from '../../support/utils/PipelineUtils';
+
+const adapterName = 'simulator';
+
+const adapterInput = AdapterBuilder.create('Machine_Data_Simulator')
+    .setName(adapterName)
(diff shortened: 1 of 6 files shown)


muyangye:
Let me actually refactor these 3 tests to a new file called `enhancedDeleteAdapter.smoke.spec.ts` so that the structure is better.

muyangye:
@bossenti, in case you did not notice, I have refactored the test. Can you trigger CI again to see if the tests pass?

muyangye:
I noticed the failed tests and fixed the problem. This time CI should not fail.

dominikriemer (approved):

ui/cypress/tests/adapter/enhancedDeleteAdapter.smoke.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+describe('Test Enhanced Adapter Deletion', () => {
+    // In the beginning, create the non-admin user
+    before(() => {
+        cy.initStreamPipesTest();
+        UserUtils.addUser(UserUtils.adapterAndPipelineAdminUser);
+        cy.logout();
+    });
+
+    // After the last test, let admin user clean up (delete the adapter, associated pipelines, and the non-admin user)
+    after(() => {
```
tenthe, line 57: Is this part of the test or is it only meant as clean up? If it is the latter, it can be removed. Before a test is started, the existing system is reset with `cy.initStreamPipesTest()`. If there were still StreamPipes resources in your tests, we should delete them. The code for the reset can be found in the class `ResetResource.java`. 
This ensures that all tests always have a clean environment, even if the previous test fails.

tenthe (changes requested):
Thanks for the PR. 
I had one comment regarding the clean up of the test. 
Let's discuss it :)

ui/cypress/tests/adapter/enhancedDeleteAdapter.smoke.spec.ts
muyangye: This is only meant as clean up. And thanks for the reminder as I just saw that there's such a class. However, [ResetManagement's reset method](https://github.com/apache/streampipes/blob/a6d219a4c8f97c23e328454ed2cdc5da9c87c33f/streampipes-rest/src/main/java/org/apache/streampipes/rest/ResetManagement.java#L58C11-L58C11) only deletes all adapters, data lakes, and visualizations, whereas for this test, a new user is created. If the user is not deleted, then later E2E tests under `userManagement` will fail. I guess the only solution here would be to only delete the user in the clean up. What do you think?

tenthe: Hi @muyangye,

I would say we need to update the `org.apache.streampipes.rest.impl.ResetResource` endpoint to remove other users as well. Any e2e test should have a clean instance with an admin user as a starting point. I have opened an issue for this #2223.

Regarding your PR, I would suggest that we remove this part because if there is a problem with the cleanup procedure, the following tests will fail. Alternatively, we can leave it in the code for now and remove it once issue #2223 is fixed.

What do you think of this suggestion?

muyangye: Oh that makes perfect sense! It's much better to have a really clean Streampipes instance every time before test. The only issue with the approach is that if we remove the cleanup code in my PR it can't be merged because later tests will fail. #2223 has to be done before merging my code. Since I believe #2223 only requires a simple change, if you don't mind solving 2 issues in 1 PR, I can implement the altogether in this PR.

tenthe: Cool, thanks. It is totally up to you. If you want, you can fix it in this PR as well. Otherwise, we can merge this PR and remove the code with the new PR of #2223.
What do you prefer?

muyangye: Let me fix both of them in this PR and change the title.

muyangye: Just fixed #2223!

tenthe (approved):
Hey @muyangye, 
thanks a lot for directly integrating the reset code

dominikriemer (approved):

Issue #2199, opened 2023-11-17
Nginx configuration parameters as environment variables


obermeier:
With the new modules _extensions-all-jvm_ and _extensions-iiot-minimal_ the nginx configuration should be updated to be able to use the HTTP-Server adapter the established way.

Since the hostname of the service could differ (extensions-all-jvm, extensions-iiot-minimal, ...), what do you think of the idea of adding an option to set the host via an environment variable. 
The current ui docker image contains the tool `envsubst` which is able to set the values in the nginx configuration.

A short example: [Link](https://github.com/obermeier/streampipes/commit/405a92950b5a21c6eca3ae4b5cd7cc302222e71b)</div>

PR #2203, opened 2023-11-18
Nginx conf parameters as env var

obermeier:
Convert example from #2199 to PR.

PR introduces (a) breaking change: no

PR introduces (a) deprecation: no


bossenti (commented):
Thanks a lot for this PR @obermeier!

I think we should also set `SP_HTTP_SERVER_ADAPTER_ENDPOINT` in our Helm chart 🙂

dominikriemer:
What happens if the variable is not provided?
Will nginx fail to start or just use the empty route?

For the helm charts, we can probably just add a route to the ingress and ingressroutes since we don't use the proxy definitions from the conf file in the k8s setup.

obermeier:
> What happens if the variable is not provided? 
 It uses the default.conf without modifictions [Link](https://github.com/apache/streampipes/blob/405a92950b5a21c6eca3ae4b5cd7cc302222e71b/ui/docker-entrypoint.sh#L22C11-L22C11)

> Will nginx fail to start or just use the empty route?
It uses ``extensions-all-jvm:8090`` [Link](https://github.com/apache/streampipes/blob/405a92950b5a21c6eca3ae4b5cd7cc302222e71b/ui/nginx_config/default.conf#L47) from default.conf.

> For the helm charts, we can probably just add a route to the ingress and ingressroutes since we don't use the proxy definitions from the conf file in the k8s setup.

I like the idea

obermeier:
> Thanks a lot for this PR @obermeier!
> 
> I think we should also set `SP_HTTP_SERVER_ADAPTER_ENDPOINT` in our Helm chart 🙂


And the installer/cli scripts

obermeier:
I added the implementations for the installers.

What would you propose in case of a migration? (The case if a configuration exists)

obermeier:
On start 

> I added the implementations for the installers.
> 
> What would you propose in case of a migration? (The case if a configuration exists)

It creates now a backup of the old configuration and applies the new configuration.
Is this solution ok?

bossenti:
@tenthe @dominikriemer Can you have a look here?
I can't assess the changes

dominikriemer:
Hi @obermeier sorry this took me so long.
I think the changes look ok, the only question I have is in the k8s docs - are we re-adding the consul configs in this PR?

obermeier:
Hi @dominikriemer, thank you so much for your review. I realized there were some errors in the process of merging with the dev branch. I've now pushed an updated version of the k8s documentation.

installer/k8s/README.md
```diff
@@ -179,20 +180,7 @@ rm -rf ${HOME}/streampipes-k8s
...
-| external.couchdb.port                           | Port for the CouchDB service                             | 5984                                     |
-| external.couchdb.service.name                   | Name of the CouchDB service                              | "couchdb"                                |
-| external.couchdb.service.port                   | TargetPort of the CouchDB service                        | 5984                                     |
-| external.couchdb.persistence.storageClassName   | Storage class name for CouchDB PVs                       | "hostpath"                               |
-| external.couchdb.persistence.storageSize        | Size of the CouchDB PV                                   | "1Gi"                                    |
-| external.couchdb.persistence.claimName          | Name of the CouchDB PersistentVolumeClaim                | "couchdb-pvc"                            |
-| external.couchdb.persistence.pvName             | Name of the CouchDB PersistentVolume                     | "couchdb-pv"                             |
+
```
dominikriemer, line 183: Looks much better now, but does this delete the couchdb parameters? ;-)

obermeier: Yes, I think it exits twice in this README.md. For this reason, I removed one version.

dominikriemer: Great, thank you!

dominikriemer (approved):

IsaakKrut:
@obermeier @dominikriemer change in installer/cli/deploy/standalone/extensions-all-jvm/docker-compose.yml broke local docker-compose startup: 
\installer\cli\deploy\standalone\extensions-all-jvm\docker-compose.yml: yaml: unmarshal errors:
  line 24: mapping key "environment" already defined at line 20

If you could add the new env variable under the existing environment

obermeier:
Hi @IsaakKrut thank you very much for this hint! 
I created a PR #2203.

PR #2300, opened 2023-12-09
[#2181] Redesign FileAPI to use original file name

### LINKED ISSUE #2181: Redesign FileAPI to use original file name
### Body

As of now, every file uploaded to StreamPipes receives a unique identifier.
While this is conveniant to handle multiple files with the same name, it has some drawbacks for document retrieval. One is, that it is not possible to query a file by its original name safely.
Furthermore, it might get confusing for users if there are, e.g., five files displayed with the same name.
Therefore, we should aim to use the file name as identifier.
But we need to consider here how we can allow the user to upload a file whose file name is already in use, e.g, by opening a dialog that allows to enter a new file name or proposes an adapted file name.

### StreamPipes Committer

muyangye:
### Purpose

Try to fix #2181 

### Remarks

After this change, if the users are trying to upload a file that already exists (same original file name ignoring case):
4b6edcdc19237ebba5a6203cb9a101d
They can upload just like before after rename:
4e957bb66de0a8efb109f1836c35acc
Similarly, when users can upload multiple files and some of them already exist:
baf592794791a090bfb4dc1667a54d3
The dialog will keep appearing if even after rename, those files still exist:
9c2b2df4708ab6eff414548cdd208ac
As you may noticed, there are already some files with same name in the background. I created those files before this change to test migration. After the migration, they shall be renamed:
a16442acef371bad4aa52cf7dd69f2b
00e5b52edead535d6fdd30cddd078b6
I have also created adapters using those files and verified they still work (as well as pipelines using those adapters):
032a4cb01e70878cc8682d022a9f6aa
10b2e6666174d825dbf869cf6b4b022

I have not implemented E2E tests for this new feature because the discussion of #2181 is ongoing. If everyone is happy with this solution. I'll go ahead create an issue to implement the E2E tests.

I know the code is a little bit long. So TLDR, I have kept `internalFilename` and `originalFilename` to avoid damages to other processes using current files. However, I added an endpoint `/allFilenames` in `PipelineElementFile` that fetches all current `originalFilename`s so that in the frontend I can check if some files users uploaded already exist and if so, let users rename those files. In addition, since before this change, there are probably some files with the same name, I have implemented a `DuplicateFilesRenameMigration` to rename those files' `originalFilename`.


streampipes-rest/src/main/java/org/apache/streampipes/rest/impl/PipelineElementFile.java
```diff
@@ -119,4 +120,14 @@ public Response getFile(
...
     return ok(FileManager.getFile(filename));
   }
+
+  @GET
+  @Path("/allFilenames")
+  @Produces(MediaType.APPLICATION_JSON)
+  @PreAuthorize(AuthConstants.HAS_READ_FILE_PRIVILEGE)
+  public Response getAllOriginalFilenames() {
+    return ok(FileManager.getAllFiles().stream().map(fileMetadata -> fileMetadata.getOriginalFilename().toLowerCase())
+        .collect(
```
bossenti, line 130:
```suggestion
        .toList();
```

streampipes-rest/src/main/java/org/apache/streampipes/rest/impl/PipelineElementFile.java
```diff
@@ -119,4 +120,14 @@ public Response getFile(
...
+            Collectors.toList()));
```
bossenti, line 131:
```suggestion

```

streampipes-service-core/src/main/java/org/apache/streampipes/service/core/migrations/v095/DuplicateFilesRenameMigration.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.List;
+
+public class DuplicateFilesRenameMigration implements Migration {
+  @Override
+  public boolean shouldExecute() {
+    return true;
+  }
+
+  @Override
+  public void executeMigration() {
```
bossenti, line 37: Would be great if you could add some documentation here about why this migration is necessary. So that we easily can relate this migration to it's purpose later on as well.

streampipes-service-core/src/main/java/org/apache/streampipes/service/core/migrations/v095/DuplicateFilesRenameMigration.java
```diff
@@ -0,0 +1,68 @@
...
+    var duplicateFiles = new HashMap<String, List<FileMetadata>>();
+    var fileMetadataStorage = StorageDispatcher.INSTANCE.getNoSqlStore().getFileMetadataStorage();
+    for (var metadata : StorageDispatcher.INSTANCE.getNoSqlStore().getFileMetadataStorage()
```
bossenti, line 40:
```suggestion
    for (var metadata : fileMetadataStorage
```

ui/src/app/files/dialog/file-rename/file-rename-dialog.component.html
```diff
@@ -0,0 +1,70 @@
...
+
+<div class="sp-dialog-container">
+    <div class="sp-dialog-content p-15">
+        <div fxFlex="100">
+            <div fxFlex="100" style="margin: 5px; width: 100%">
+                <div fxLayout="column" style="width: 100%">
+                    <div fxFlex="100" fxLayoutAlign="center" fxLayout="column">
+                        <b>
+                            <h4>
+                                The file uploaded already exists, please choose
```
bossenti, line 27:
```suggestion
                                A file with the same name already exists. Please give your file a new name or select one of the existing files.
```

bossenti (changes requested):
Thank you very much for this PR and the detailed documentation @muyangye. I really appreciate!
I've left some minor remarks for improvements.

@dominikriemer Can you give the UI changes a review?

streampipes-service-core/src/main/java/org/apache/streampipes/service/core/migrations/v095/DuplicateFilesRenameMigration.java
muyangye: That's a good suggestion! I'll add some documentation.

streampipes-rest/src/main/java/org/apache/streampipes/rest/impl/PipelineElementFile.java
muyangye: Hmmm, don't know Java now has this nice function!

tenthe (approved):
Hi @muyangye,
thanks a lot for your PR.  I think it's very good.
I have created another branch myself in which I have written a unit test for the DuplicateFilesRenameMigrationTest.executeMigration() method. I also refactored it a bit. What do you think? Do you think this iteration is an improvement? [Link to branch](https://github.com/apache/streampipes/commit/830ba8f6f08f911dffe977cdbecac0a4010cd815)

muyangye:
> Hi @muyangye,
> thanks a lot for your PR. I think it's very good.
> I have created another branch myself in which I have written a unit test for the DuplicateFilesRenameMigrationTest.executeMigration() method. I also refactored it a bit. What do you think? Do you think this iteration is an improvement? [Link to branch](https://github.com/apache/streampipes/commit/830ba8f6f08f911dffe977cdbecac0a4010cd815)

Hi @tenthe, thanks a lot for the refactoring! I think those code are much more self-explanatory! Thanks for the unit tests too. I didn’t write unit tests because this migration is a temporary solution. I’ll come up with unit tests for my next PR. Btw I still haven’t added a comment that Tim suggested and I will do that once I land (checking in for a flight right now) :)

muyangye:
@bossenti @tenthe Sorry for the late reply. I am currently in China and I was trying to solve Github being blocked for the past 2 days and found no luck:

I used a VPN which is needed to even access Github but still couldn't do git push. I just realized I could directly commit on web. Can you merge your branch since I can't do anything with Git @tenthe? Sorry for the inconvenience...

muyangye:
Closing this PR as updated by #2356.

PR #2311, opened 2023-12-12
Extend all MS Teams sink by a silent period

bossenti:
### Purpose

As of now, only the Email and internal notification sink have the option to configure a silence period which avoids spamming the target.
This feature is really valuable and is therefore added to all notification sinks in StreamPipes. This is achieved by introducing a shared parent class called `StreamPipesNotificationSink` that contains everything related to the silence period.
So when implementing a notification sink, one does not have to account for the silence period in any way. Such a implementation looks like the following:

```java
public class DemoSink extends StreamPipesNotificationSink {
  
  @Override
  public void onInvocation(
      SinkParams parameters,
      EventSinkRuntimeContext runtimeContext
  ) throws SpRuntimeException {
    super.onInvocation(parameters, runtimeContext);
    
    //...
  }
  
  @Override
  public void onDetach() {
    
  }

  @Override
  public void onNotificationEvent(Event inputEvent) {

  }

  @Override
  public DataSinkBuilder declareModelWithoutSilentPeriod() {
    return null;
  }
}
```

### Remarks


tenthe:
Hi @bossenti,
the PR looks good. I have a question about migrating sinks that we had in previous releases. Do we need a migration script for them because there is a new configuration option?

bossenti:
Indeed!
Thanks a lot for reminding @tenthe 

I'll add the corresponding migrations

streampipes-extensions/streampipes-sinks-internal-jvm/src/main/java/org/apache/streampipes/sinks/internal/jvm/notification/InternalStreamPipesNotificationSink.java
```diff
@@ -61,63 +53,59 @@ public class NotificationProducer extends StreamPipesDataSink {
...
+                           .toString());
+    notification.setRead(false);
+    notification.setTitle(title);
+    notification.setMessage(PlaceholderExtractor.replacePlaceholders(inputEvent, content));
+    notification.setCreatedAt(currentDate);
+    notification.setCreatedAtTimestamp(currentDate.getTime());
+    notification.setCorrespondingPipelineId(correspondingPipelineId);
+    notification.setTargetedAt(correspondingUser);
+
+    // TODO add targeted user to notification object
```
tenthe, line 84: Can we remove this and create an issue for this feature instead?

tenthe (approved):

streampipes-extensions/streampipes-sinks-internal-jvm/src/main/java/org/apache/streampipes/sinks/internal/jvm/notification/InternalStreamPipesNotificationSink.java
bossenti: Sure, here it is https://github.com/apache/streampipes/issues/2314

bossenti:
@tenthe I removed the silent period from existing sinks and created an issue for their adoption: https://github.com/apache/streampipes/issues/2312
This should keep this PR small

dominikriemer (approved):

Issue #2374, opened 2023-12-27
Provide a framework to easily unit test processing elements


tenthe:
## Description
It is currently challenging for developers to write unit tests for processing elements. A dedicated test framework is needed to facilitate testing with various user inputs efficiently. The focus should be on extensive testing of the onEvent method.

## Proposed API
``` java
var processor = new ProcessingElementUnderTest();
 
var userConfiguration = Map.of("key-textFIlter", "text");
 
var inputEvents = List.of(
     new Event(),
     new Event()
 );
 
var expectedOutputEvents = List.of(
     new Event(),
     new Event()
 );
 
ProcessingElementTestExecuter.run(processingElement, userConfiguration, inputEvents, expectedOutputEvents);
```

### StreamPipes Committer

I acknowledge that I am a maintainer/committer of the Apache StreamPipes project.

PR #2375, opened 2023-12-27
feat(#2374): Add suggestion to test processing elements

tenthe:
### Purpose

The current pull request introduces a prototype for unit testing processing elements in the `streampipes-processors-filter-jvm` module, specifically for the `NumericalFilterProcessor`.

#### Proposal
- Implemented a basic prototype allowing developers to test processing element configurations.
- Used `Map<String, Object>` for user input and `List<Map<String, Object>>` for events.
- Proposing a discussion on whether to keep the current data type approach for simplicity or consider alternatives.
- A static class, `ProcessingElementTestExecutor`, is suggested to execute the actual tests.

> [!IMPORTANT] 
> This PR is not final; its purpose is to showcase the testing functionality and initiate a discussion on testing options.

If you have any suggestions or opinions on the proposed approach, feel free to share them here.

### Remarks


IsaakKrut:
@tenthe do you have examples of tests or processors that use selector prefix or event schema? Are there any that use multiple prefixes? Trying to understand which processors require something extra for testing

tenthe:
Hi @IsaakKrut,

my suggestion would be that you try to write a test for another processor with this "new" API.
And then, based on this experience, we discuss and try to design the API and the implementation for the "ProcessingElementTestExecutor" together. 
What do you think of this idea?

I think the biggest challenge will be to handle all the different types of static properties that a user can enter to configure the processor. I think one or two more sample tests would help to get a better understanding of what the best solution should look like.

Cheers,
Philipp

tenthe:
> Hi @tenthe @IsaakKrut,
> 
> I would love to contribute to this effort. I have already migrated the classes stated in #2538 to JUnit 5, I am quite unclear on what needs to be done for my work to integrate into this effort.
> 
> Thanks, Pambrocio

Hi @pambrocio,
thanks a lot for your PRs. I do not have much time this week, I'll have a detailed look at them next week.

Cheers,
Philipp

tenthe:
Hi @IsaakKrut,
I have fixed the problem with the cyclic dependencies. I also fixed the checkstyle issues.

Is it possible for you to extract the `TestConfigurationBuilder` into an own file?

@pambrocio, I suggest that we clean up this branch and then merge it into dev.
Based on these changes, you can customize your PRs as the API should now be cleaner.
And we can then use your feedback to further improve the testing API for processors.
Is that ok for you?

Cheers,
Philipp

bossenti (approved):

Issue #2144, opened 2023-11-06
Enhance Static Properties by UI validation


bossenti:
### Body

Currently, user input that is collected via the UI when configuring adapters & pipeline elements is not validated, e.g., for being an integer.
This can lead to frustration for the end user.
Therefore, we should try to build in validation for our static properties so that at least data types are validated automatically and maybe even additional validation criteria can be defined.

### StreamPipes Committer

I acknowledge that I am a maintainer/committer of the Apache StreamPipes project.

IsaakKrut:
Hi @bossenti, I looked at the adapters creation page and saw some validation is already in place. For example: 
![image](https://github.com/apache/streampipes/assets/25233902/914c3e24-e508-433e-93fb-49f4daaf0be8)

Could you please clarify what properties you are referring to and where they can be found?

tenthe:
Hi @IsaakKrut,
can you please check if the validation also validates that the input is an integer?
Is there an error message when you enter "9092.0"?

IsaakKrut:
@tenthe it does not, but it shows an error with exception stack trace when it tries to connect and guess the schema. I did find 3 adapters that do not have UI validation on their port fields: OPC UA, PLC4X MODBUS, ROS Bridge

tenthe:
Ok great. Then this is a good starting point for this issue

PR #2401, opened 2024-01-07
#2144 Enhance Static Properties by UI validation

IsaakKrut:
Updated port number fields to Integer type for following adapters:
Ros Bridge
OPC UA
PLC4X MODBUS

Remarks


tenthe:
Hi @IsaakKrut,
thanks a lot for submitting this pull request. I think this is a great improvement for users.

I was wondering if this change might require a migration script, as it will affect existing adapters during system upgrades.

Therefore, I would recommend providing an update script for the affected adapters. An example can be found in [1]. This change affected the PLC4X S7 adapter, where an alternative static property was added.

In general, the update script should ensure that the changes are correctly propagated to the user configurations of the existing adapter instances. In your case, the `requiredDatatype` of the `FreeTextStaticProperty` must be changed from `Datatypes.String` to `Datatypes.Integer` within the update script.

If you have any questions or need help with the implementation, please feel free to ask.

[1] https://github.com/apache/streampipes/blob/dev/streampipes-extensions/streampipes-connectors-plc/src/main/java/org/apache/streampipes/extensions/connectors/plc/adapter/migration/Plc4xS7AdapterMigrationV1.java

IsaakKrut:
@tenthe Added the migration scripts. I was not able to test it locally, is there a way to set up different adapters for testing?

tenthe:
Thank you for providing the migrations.
Unfortunately, I do not currently have a setup to test them locally. 

However, we  need a way to test such migrations. Perhaps we can create unit tests for this? What do you think? 
To test it e2e you would have to setup a system with the old version, update the system to the new dev version and check if it worked. But I am not sure how we could automate this.

At least I have added the label migration for the PR. Maybe we can use this in the future to label all PRs that contain a migration to test them again before the release.

IsaakKrut:
I added a sample unit test for one of the migrations, let me know what you think. Maybe this approach becomes more messy for elements with larger configurations as it all has to be moved to those static methods

dominikriemer (approved):

tenthe:
Many thanks for the PR @IsaakKrut.
Sorry for the late reply.

The test looks good.
@dominikriemer I was wondering what happens to the existing user input during configuration.
The same migration is used for both, the description of the processing elements and for the instances, right?

So I guess we need to check if a value is set in the original `StaticProperty` and set it for the new `StaticProperty` as well.
For the `RosBridgeAdapterMigrationV1` this would mean that we add the 'original' value to the `FreeTextStaticProperty` created in line 52.

Cheers,
Philipp

tenthe:
Hi @IsaakKrut,

I have resolved the concerns regarding the values of the static properties for the OPC UA Adapter. Therefore, I have included these updates in pull request #2443.

Please review these changes and make appropriate adjustments to the remaining adapters.
Do the changes work from your point of view?

Cheers,
Philipp

PR #2430, opened 2024-01-21
[#2181] Merge internalFilename and originalFilename

muyangye:
### Purpose
See #2181, finally merging internalFilename and originalFilename to one.

### Remarks
Result:
I created 3 files with the same name using StreamPipes 0.93.0 release:
14ed46d5129d58af83ba1c22e0a0d52
After which I run the latest migration which is successful:
4ebe269d0724a23570fa49df7d94a90
However, for **pipelines, specifically Data Lake (not adapters, adapters can produce messages as usual (i.e. if I create another pipeline using the same adapter created in v0.93.0, I can still visualize it correctly))** created before using File Stream, DataLake expects a new schema:
508237bb823499217ffdc5b9a4285cc
After which all pipelines are back to normal:
84b5655d8abfbdaa62050342c26de28

Method:
Get raw json stored in CouchDB => use file ID as the "link" between old version of `FileMetadata` and new version of `FileMetadata` => then update `FileMetadata` and File stored locally.

Test:
I haven't written any unit tests yet because I don't know how to test it since `FileMetadata` class was changed. My original thought is this: create a new class called `oldFileMetadata` which is the same as the `FileMetadata` class that has both originalFilename and internalFilename => store that object using a `CouchDbClient` in the FileMetadata database => run the migration. However, I don't understand why but `Utils.getCouchDbFileMetadataClient()`throws an exception of "unknown host" in file I created `streampipes-service-core/src/test/java/org.apache.streampipes.service.core.migrations.v095/MergeFilenamesAndRenameDuplicatesTest.java` so I deleted it. I believe it's a package/config issue, do you have any idea why? Happy to write a unit test after fixing this issue and everyone agrees with my approach :)

PR introduces (a) breaking change(s): yes? Please let me know if the required manual schema update of Data Lake is a problem.


dominikriemer (approved):

tenthe:
@muyangye thank you for updating the PR.
I think your approach looks good. 
I just had a couple of small comments.

Regarding the exception in the test, I guess you have to mock the static method somehow, otherwise it tries to connect to the couchdb.

muyangye:
@tenthe I will try to see if I can mock the static method somehow. Thanks for pointing out!

Did you leave some small comments on code? For some reason I did not see them?

tenthe (changes requested):

tenthe:
> @tenthe I will try to see if I can mock the static method somehow. Thanks for pointing out!
> 
> Did you leave some small comments on code? For some reason I did not see them?

Sorry, they were still on pending. Can you now see them?

streampipes-model/src/main/java/org/apache/streampipes/model/file/FileMetadata.java
```diff
@@ -28,15 +28,16 @@ public class FileMetadata {
...
-  private String originalFilename;
+  private String filename;
   private String filetype;
 
   private long createdAt;
   private long lastModified;
 
   private String createdByUser;
 
+  public String test;
```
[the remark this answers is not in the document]
muyangye: Yes, this is solely for testing purposes to see what will happen if I add another attribute to the FileMetadata class, will delete.

streampipes-pipeline-management/src/main/java/org/apache/streampipes/manager/file/FileHandler.java
```diff
@@ -39,6 +40,16 @@ public File getFile(String filename) {
     return FileUtils.getFile(makeFile(filename));
   }
 
+  public void renameFile(String oldFilename, String newFilename) {
+    try {
+      var fileInputStream = new FileInputStream(getFile(oldFilename));
+      deleteFile(oldFilename);
+      storeFile(newFilename, fileInputStream);
+    } catch (Exception e) {
+      System.out.println(e.getMessage());
```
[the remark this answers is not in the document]
muyangye: Both are definitely doable, I will push my changes together with the unit tests.

muyangye:
@tenthe I have pushed an update that addressed your code review and added unit tests. I admit that the code is really ugly given that I added an `isTesting` field in the migration and Java doesn't support default parameters. Do you have any idea that can decouple the migration and its test other than recreating all the methods?

muyangye:
Hi @tenthe, did you have a chance to review my unit tests?

tenthe (approved):
@muyangye sorry for the late review. 
Your changes look good. Thanks a lot

streampipes-service-core/src/main/java/org/apache/streampipes/service/core/migrations/v095/MergeFilenamesAndRenameDuplicatesMigration.java
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  /**
+   * Converts InputStream (as stored in CouchDB) to Map, if there's an error, constructs a new Map
+   */
+  private Map<String, Object> convertInputStreamToMap(InputStream inputStream) {
+    try {
+      return mapper.readValue(inputStream, Map.class);
+    } catch (Exception e) {
+      return new HashMap<>();
```
bossenti, line 114: Can we please add a log message here?
This allows for better debugging in case we have some problems with the migration

bossenti (commented):
@muyangye Thank you very much for your dedicated work!
I have only two last minor remarks

streampipes-service-core/src/main/java/org/apache/streampipes/service/core/migrations/v095/MergeFilenamesAndRenameDuplicatesMigration.java
muyangye: Just added a log message. By the way even if this exception occurs, rest of the files' migration will not be affected since it will just be skipped.

bossenti (approved):

ui/cypress.config.ts
```diff
@@ -44,7 +44,7 @@ export default defineConfig({
             return require('./cypress/plugins/index.ts')(on, config);
         },
         specPattern: 'cypress/tests/**/*.{js,jsx,ts,tsx}',
-        baseUrl: 'http://localhost:80',
+        baseUrl: 'http://localhost:8082',
```
bossenti, line 47: I assume this was committed by accident, right?

bossenti (changes requested):

ui/cypress.config.ts
muyangye: Ah yes, I accidentally did "git commit -am" instead of only committing one file. I have reverted all the irrelevant changes. Thank you for catching it!

bossenti (approved):

bossenti:
Thank you very much for your contribution here @muyangye 🙏🏼

PR #2443, opened 2024-01-29
#2144 Enhance Static Properties by UI validation V2

tenthe:
### Remarks


tenthe:
@IsaakKrut, there seems to be an issue with the unit test. Do you have an idea what is going wrong. 
I tried to mock the extractor but it did not work.
Cheers,
Philipp

IsaakKrut:
@tenthe The 3 values(serverAddressValue, hostValue, portValue) from the extractor you are using in OpcUaAdapterMigrationV2 have to be present in the extractor. Without them I'm getting NullPointerException in the AbstractParameterExtractor.singleValueParameter method call. Could we instantiate an extractor with those properties inside the test? And if so, are they gonna be available during actual migration as well?

tenthe:
@IsaakKrut, I tested the migration and there I did not get any exceptions.
I only geth the NullPointer when running the test. Maybe we can mock the extractor return a value when the function is called with the parameters (serverAddressValue, hostValue, portValue).
Do you think this will work?

IsaakKrut:
@tenthe All these 3 fields (serverAddressValue, hostValue, portValue) are going to have a value during migration, right? If so we can add the defaultValue parameter for host and port for version 2 config in OpcUaAdapterVersionedConfig on lines 138, 140. And create an extractor inside the test with the same default values. Ran the test with those changes and it passed
image

image

tenthe:
Ok, great. Can you add your changes. Then we can close this PR.

IsaakKrut:
Created a separate PR to update this branch since I don't have push access

#2454

IsaakKrut:
Hi @tenthe, did you have a chance to look at PR #2454?

tenthe:
Hi @IsaakKrut,
sorry for the late reply. I'll have a look at it in the upcominig days.
Cheers,
Philipp

IsaakKrut:
Hi @tenthe,

Thanks for the update, looks much cleaner. Both the implementation and the tests. I tested locally and it works as expected

dominikriemer (approved):
Looks good to me. Thanks a lot @IsaakKrut and sorry that this took so long!

PR #2474, opened 2024-02-18
[#2473] Nginx configuration migraton to version 0.93.0+

### LINKED ISSUE #2473: Nginx configuration error during a migration from version 0.93.0 to 0.95.0.

Backend, UI

### What happened?

In version 0.95.0, the Nginx configuration could be generated based on a template file. 
However, this template file is not accessible in the filesystem if the underlying volume was previously used for another instance. 
As a result, Nginx fails to start due to the missing configuration.

Workaround:
A workaround for this issue is copy the missing template file to the correct location from a second locaton (/etc/opt/nginx/). 
This step is only necessary when migrating from version 0.93.0 to a newer version. 
It is advisable to remove this workaround after the release of version 0.95.0 to maintain cleaner code.

What do you think of this ideas?

### How to reproduce?

Start StreamPipes 0.93.0 and than start 0.95.0-SNAPSHOT with the same volumes.

### Expected behavior

_No response_

_No response_

obermeier:
### Purpose
Fix Nginx startup problems because of missing configuration in case of a auto generated configuration.
 Copy the missing template file to the correct location from a second locaton (/etc/opt/nginx/)

### Remarks
Could be removed in future versions.

PR introduces a breaking change: no
PR introduces a deprecation: no

fixes  #2473

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -60,6 +60,8 @@ services:
     depends_on:
       - couchdb
       - backend
+    volumes:
+      - nginx:/etc/nginx/
     logging: *default-logging
     networks:
       spnet:
@@ -141,6 +143,7 @@ volumes:
   influxdb:
   influxdb2:
   backend:
+  nginx:
 
 networks:
   spnet:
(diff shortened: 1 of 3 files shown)


bossenti:
Hey @obermeier,

sorry for reacting that late to your PR. Could you please give some more context to the changes that have been made? I can't really assess them, to be honest 😅

bossenti (approved):
Thank you very much for the clarification @obermeier!

tenthe (approved):

PR #2502, opened 2024-03-04
Code cleanup

selutin99:
### Purpose

* Use constants if it applicable
* Remove unnecessary diamond operators
* Use try-with-resources

### Remarks


tenthe (approved):
@selutin99 thanks for the code cleanup. Your changes look good.

PR #2512, opened 2024-03-06
fix: extend list of forbidden characters in Data Explorer with "-"

bossenti:
### Remarks

### DIFF
--- a/streampipes-data-explorer-commons/src/main/java/org/apache/streampipes/dataexplorer/commons/DataExplorerUtils.java
+++ b/streampipes-data-explorer-commons/src/main/java/org/apache/streampipes/dataexplorer/commons/DataExplorerUtils.java
@@ -15,6 +15,24 @@
  * limitations under the License.
  *
  */
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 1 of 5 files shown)


tenthe (approved):

tenthe (changes requested):
We should check whether this will damage older installations before we merge this PR. 
In particular, we need to check if the changes are in the right sanitizer. I think we should move them to the `InfluxNameSanitizer`. However, this has a slightly different functionality, so it needs to be customized.

To summarize:
- We should make sure that the change does not break any old installations.
- The runtime names of the properties must be identical in the event schema and in the event.

bossenti:
Thank you Philipp for pointing this out.

So in the end, the replacement should happen in `InfluxStore.sanitizeRuntimeNamesInEvent` and `DataExplorerUtils.sanitizeDataLakeMeasure` at the same time, right?
Then we need to adapt these methods so that they not only look for reserved keywords but also for specific characters.

Do we have a chance to introduce this not as a breaking change?

dominikriemer:
Hi
I think it's ok to rename the fields/runtime names if they are not working anyways when using a `-` character.
Changing the measure name is something we should avoid since there is no way in Influx to rename an existing measurement, so this would break existing measurements which are already using this character. 
Using a `-` within a measurement name doesn't cause any problems on the writing or querying side, so there is no need to change it?

bossenti:
> Using a - within a measurement name doesn't cause any problems on the writing or querying side, so there is no need to change it?

Indeed, I was on the wrong path here

PR #2515, opened 2024-03-06
2312 Extend Notification sinks by silent period

IsaakKrut:
Purpose

For the following sinks: OneSignalSink, SlackSink, TelegramSink

Refactored to extend StreamPipesNotificationSink interface
Implemented migrations with the use of a common interface INotificationDataSinkMigrator

Added a unit test for INotificationDataSinkMigrator

Remarks

### DIFF
--- a/streampipes-extensions/streampipes-sinks-notifications-jvm/src/main/java/org/apache/streampipes/sinks/notifications/jvm/slack/SlackNotificationSink.java
+++ b/streampipes-extensions/streampipes-sinks-notifications-jvm/src/main/java/org/apache/streampipes/sinks/notifications/jvm/slack/SlackNotificationSink.java
@@ -111,8 +111,8 @@ public class SlackNotificationSink extends StreamPipesNotificationSink {
   }
 
   @Override
-  public void onNotificationEvent(Event inputEvent) {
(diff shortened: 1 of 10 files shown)


streampipes-extensions/streampipes-sinks-notifications-jvm/src/main/java/org/apache/streampipes/sinks/notifications/jvm/slack/SlackNotificationSink.java
```diff
@@ -56,10 +55,9 @@ public class SlackNotificationSink extends StreamPipesDataSink {
   private String originalMessage;
 
   @Override
-  public DataSinkDescription declareModel() {
-
+  public DataSinkBuilder declareModelWithoutSilentPeriod() {
     return DataSinkBuilder
-        .create("org.apache.streampipes.sinks.notifications.jvm.slack", 0)
+        .create("org.apache.streampipes.sinks.notifications.jvm.slack", 1)
```
dominikriemer, line 60: Does it make sense to extract the ID to a public static variable so that it can be reused in the Migration classes?

bossenti (approved):
Thank you very much for this contribution @IsaakKrut 🙏🏼

PR #2521, opened 2024-03-08
#2442 fixed data explorer time range buttons

### LINKED ISSUE #2442: Buttons in time range selection not functioning
### Body

The next and previous buttons within the time range selection of the data explorer are unresponsive. Clicking on either button does not trigger any action.

### StreamPipes Committer

IsaakKrut:
Fixes #2442 

### DIFF
--- a/ui/cypress/tests/datalake/timeRangeSelectors.spec.ts
+++ b/ui/cypress/tests/datalake/timeRangeSelectors.spec.ts
@@ -20,12 +20,12 @@ import { DataLakeUtils } from '../../support/utils/datalake/DataLakeUtils';
 
 describe('Test Time Range Selectors in Data Explorer', () => {
     const periods = [
-        '15_min',
-        '1_hour',
-        '1_day',
-        '1_week',
-        '1_month',
-        '1_year',
+        { selector: '15_min', offset: 15 },
+        { selector: '1_hour', offset: 60 },
+        { selector: '1_day', offset: 1440 },
+        { selector: '1_week', offset: 10080 },
+        { selector: '1_month', offset: 43200 },
+        { selector: '1_year', offset: 525600 },
     ];
 
     const dateAttribute = 'ng-reflect-date';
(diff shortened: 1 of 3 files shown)


IsaakKrut:
Is there a reason why 1 month offset was 60*24*30 + 600 minutes? Updated it to be exactly 30 days

ui/cypress/tests/datalake/timeRangeSelectors.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+import { DataLakeUtils } from '../../support/utils/datalake/DataLakeUtils';
+
+describe('Test Time Range Selectors in Data Explorer', () => {
+    const periods = [
```
tenthe, line 22: You could use a list of objects here instead of a list of stings, then you can remove the switch block in `getExpectedStartDate`

tenthe (commented):
Hi @IsaakKrut,
Thank you for your PR. I have checked and it worked as expected. 
As for your question about the one month delay, I'm not sure why that was the case. Your change makes perfect sense to me.

I have some minor comments for possible code improvements what do you think about them?

I really like the fact that you have directly provided an e2e test for this function. For me the test was a bit hard to read, is it possible to refactor it a bit to make the individual steps clearer.
One reason I found it difficult may be that I am not familiar with the Cypress `invoke` keyword. How does it work exactly?

Thanks again for your contribution.

IsaakKrut:
Hi @tenthe,

Updated the test with your suggestions and fixed a daylight savings related issue. 

Regarding invoke method. What it does is it invokes a function specified in the first parameter on an object yielded before it in the chain. In this case, it's an attr() function that gets the attribute's value based on the second parameter which is attributeName

This is the html element:
`<sp-date-input data-cy="time-range-to" _nghost-ng-c1192185161="" ng-reflect-date="Mon Mar 11 2024 16:23:16 GMT-0"><input _ngcontent-ng-c1192185161="" type="datetime-local" step="1" class="sp-date-input"></sp-date-input>`

So the current date selection is stored in the "ng-reflect-date" attribute. I added a couple of comments, let me know if I can do anything else to make it easier to read

PR #2522, opened 2024-03-09
refactor(#2520): Add method to check file content in FileProtocolUtils

### LINKED ISSUE #2520: Issue with File Cache Persistence in Extensions Service
### Body

- Files are being cached within the extensions service as expected.
- However, when a user deletes a file (such as a file stream adapter), it's correctly removed from the backend but remains in the extensions service cache.
- Consequently, when a new file with the same name is uploaded, the extension service incorrectly uses the old cached file instead of the updated file.

This behavior is also the reason for the failing nightly e2e tests.

### StreamPipes Committer

tenthe:
### Purpose

Fix #2520 

### Remarks

### DIFF
--- a/streampipes-pipeline-management/src/main/java/org/apache/streampipes/manager/file/FileManager.java
+++ b/streampipes-pipeline-management/src/main/java/org/apache/streampipes/manager/file/FileManager.java
@@ -17,19 +17,17 @@
  */
 package org.apache.streampipes.manager.file;
 
+import org.apache.streampipes.commons.file.FileHasher;
 import org.apache.streampipes.model.file.FileMetadata;
 import org.apache.streampipes.sdk.helpers.Filetypes;
 import org.apache.streampipes.storage.api.IFileMetadataStorage;
 import org.apache.streampipes.storage.management.StorageDispatcher;
 
-import org.apache.commons.codec.digest.DigestUtils;
 import org.apache.commons.io.input.BOMInputStream;
 
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
-import java.nio.file.Files;
-import java.nio.file.Paths;
 import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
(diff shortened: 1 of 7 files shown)


streampipes-pipeline-management/src/main/java/org/apache/streampipes/manager/file/FileManager.java
```diff
@@ -104,6 +107,17 @@ public static InputStream cleanFile(InputStream fileInputStream, String filetype
     return fileInputStream;
   }
 
+  public static boolean checkFileContentChanged(String filename, String hash) throws IOException {
+    var fileHash = getFileHash(filename);
+    return !fileHash.equals(hash);
+  }
+
+  private static String getFileHash(String filename) throws IOException {
```
bossenti, line 115: This is required to be the same implementation as in `FileProtocolUtils`.
Do you see a way where we can define the method only once?
If this is not possible, I would like to some notes that make this interdependence explicit.

streampipes-rest/src/main/java/org/apache/streampipes/rest/impl/PipelineElementFile.java
```diff
@@ -139,4 +139,24 @@ public ResponseEntity<List<String>> getAllOriginalFilenames() {
...
+  @PreAuthorize(AuthConstants.HAS_READ_FILE_PRIVILEGE)
+  public ResponseEntity<Boolean> checkFileContentChanged(
+      @PathVariable(value = "filename") String filename,
+      @PathVariable(value = "hash") String hash
+  ) {
+    try {
+      return ok(FileManager.checkFileContentChanged(filename, hash));
+    } catch (IOException e) {
+      throw new SpMessageException(
+          org.springframework.http.HttpStatus.NOT_FOUND,
```
bossenti, line 155: Is there a reason why you don't import `HttpStatus`?

bossenti (approved):

PR #2547, opened 2024-03-13
Use pydantic 2.x dependency

obermeier:
### Purpose
It seems to me that pydantic 2.x gets more and more popular. 
What are your thoughts on upgrading the dependency from version 1.x to 2.x to prevent (more) conflicts with other libraries?

PR introduces (a) breaking change(s): <no>

PR introduces (a) deprecation(s): <no>


bossenti:
Hey @obermeier,

strong +1 from my side.
I was just short of time to do the migration so far but I'm completely fine with your approach.
Thank you for bringing this up 🙏🏼

bossenti (approved):

dominikriemer (approved):

PR #2559, opened 2024-03-14
fix(#2549): consider renaming rules for timestamp extraction

### LINKED ISSUE #2549: Property renaming in the FileStream Adapter does not work

Processing Elements

### What happened?

I encountered an issue while utilizing the File Stream Adapter in StreamPipes. Within my workflow, I needed to rename certain properties, particularly focusing on the timestamp property. However, despite following the prescribed steps for property renaming, the renaming process does not take effect.

### How to reproduce?

_No response_

### Expected behavior

_No response_

_No response_

None

bossenti:
### Purpose

fix #2549

When the timestamp property is renamed, this breaks the current timestamp extraction in the FileReplayAdapter since we are looking for the runtime name.
Therefore, we need to account for renaming rules when extracting the timestamp from the source data with the correct key.

### Remarks

### DIFF
--- a/streampipes-extensions/streampipes-connect-adapters-iiot/src/main/java/org/apache/streampipes/connect/iiot/protocol/stream/FileReplayAdapter.java
+++ b/streampipes-extensions/streampipes-connect-adapters-iiot/src/main/java/org/apache/streampipes/connect/iiot/protocol/stream/FileReplayAdapter.java
@@ -32,6 +32,7 @@ import org.apache.streampipes.extensions.management.connect.adapter.parser.JsonP
 import org.apache.streampipes.extensions.management.connect.adapter.parser.xml.XmlParser;
 import org.apache.streampipes.model.AdapterType;
 import org.apache.streampipes.model.connect.guess.GuessSchema;
+import org.apache.streampipes.model.connect.rules.schema.RenameRuleDescription;
 import org.apache.streampipes.sdk.StaticProperties;
 import org.apache.streampipes.sdk.builder.adapter.AdapterConfigurationBuilder;
 import org.apache.streampipes.sdk.helpers.Alternatives;
@@ -48,6 +49,8 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
+import java.util.List;
+import java.util.Optional;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


streampipes-extensions/streampipes-connect-adapters-iiot/src/main/java/org/apache/streampipes/connect/iiot/protocol/stream/FileReplayAdapter.java
```diff
@@ -133,66 +155,113 @@ public void onAdapterStarted(IAdapterParameterExtractor extractor,
...
+                } catch (InterruptedException e) {
+                  LOG.info("File stream adapter was stopped, the current replay is interuppted", e);
+                }
+              }
+            }
+
+            timestampLastEvent = actualEventTimestamp;
+            if (replaceTimestamp) {
+              // we can directly use the desired runtime name for the timestamp instead of the original here
+              event.put(timestampRuntimeName, System.currentTimeMillis());
```
tenthe, line 260: Have you checked how this affects the `Rename timestap` rule? Looking at the `RenameTransformationRule` rule, I'm not sure if the new field `timestamp` gets the value `null` when `replaceTimestamp` is active

streampipes-extensions/streampipes-connect-adapters-iiot/src/main/java/org/apache/streampipes/connect/iiot/protocol/stream/FileReplayAdapter.java
```diff
@@ -133,66 +155,113 @@ public void onAdapterStarted(IAdapterParameterExtractor extractor,
...
 
     if (timestampField.isEmpty()) {
       throw new AdapterException("Could not find a timestamp field in event schema");
     } else {
-      timestampRuntimeName = timestampField.get().getRuntimeName();
+      timestampRuntimeName = timestampField.get()
+          .getRuntimeName();
+    }
+
+    // check for renaming rules that affect the timestamp property
```
tenthe, line 169: Can you extract this functionality into a method? That could clean up the code a bit.

tenthe (changes requested):

streampipes-extensions/streampipes-connect-adapters-iiot/src/main/java/org/apache/streampipes/connect/iiot/protocol/stream/FileReplayAdapter.java
bossenti: I'll have a look

bossenti: I think it won't be `null` but it will probably be overwritten with the old value.
I'll change it.

tenthe (approved):

Issue #2337, opened 2023-12-14
Can't access "Connect" with User Role "Connect Admin"


Marcelfrueh:
### Apache StreamPipes version

dev (current development state)

### Affected StreamPipes components

Backend

### What happened?

I might have encountered a bug while testing the user roles in Configuration > Security. After adding a test adapter, I wanted to try out the different user roles. Therefore, I added a new user with the "connect admin" role. After logging in with this account, I couldn't access "Connect" in the navigation menu. Clicking on "Connect" didn't trigger anything, and I couldn't access the adapters in any other way. Similar tests with the "pipeline admin" or "dashboard admin" worked just fine.

![Bildschirmfoto 2023-12-14 um 16 47 58](https://github.com/apache/streampipes/assets/78954450/b877f1ce-cefb-4939-b4cf-e292cd4794ab)

### How to reproduce?

-Go to connect and add new Machine Data Simulator
-Go to configuration < security
-add new user with "user@streampipes.apache.org" and add role "Connect Admin"
-enable the user and save
-log out
-log in with new user
-can't access connect

### Expected behavior

As a connect admin, the user should be able to access connect.

### Are you willing to submit a PR?

None

PR #2566, opened 2024-03-15
2337 cant access connect with user role connect admin

tenthe:
### Purpose

Fix: #2337.
We had a problem with the matching of routes.
If the module route is defined in the `MODULE_NAME.module.ts`, it could happen that the route could not be resolved correctly. To solve the problem, I had to move the route names to the file `app-routing.module.ts`.
Since this file is generated automatically, the template file `appng5.module.mst` had to be changed.

### Description of the changes
- Moving the routes from `.module.ts` to `app-routing.module.ts`.
- The component `help` and its subcomponents were moved to the module `core-ui`, as otherwise the modules were not loaded correctly

### Remarks


streampipes-platform-services/src/main/java/org/apache/streampipes/ps/PipelineElementTemplateResource.java
```diff
@@ -92,32 +95,35 @@ public ResponseEntity<?> getById(@Parameter(description = "The id of the pipelin
...
+             tags = {"Pipeline Element Templates"},
+             responses = {
+                 @ApiResponse(responseCode = "200", description = "Template successfully stored")
+             })
   public ResponseEntity<Void> create(
       @RequestBody(description = "The pipeline element template to be stored",
-          content = @Content(schema = @Schema(implementation = PipelineElementTemplate.class)))
-      @org.springframework.web.bind.annotation.RequestBody PipelineElementTemplate entity) {
+                   content = @Content(schema = @Schema(implementation = PipelineElementTemplate.class)))
+      @org.springframework.web.bind.annotation.RequestBody PipelineElementTemplate entity
```
bossenti, line 105: I know it has already been there, but could you place replace this by an import statement? 🙂

streampipes-platform-services/src/main/java/org/apache/streampipes/ps/PipelineElementTemplateResource.java
```diff
@@ -149,44 +156,45 @@ public ResponseEntity<Void> delete(
...
               + "with the labels given in the pipeline element template")
-      @RequestParam("overwriteNames") String overwriteNameAndDescription,
-
+      @RequestParam(value = "overwriteNames", defaultValue = "false", required = false)
+      String overwriteNameAndDescription,
       @RequestBody(description = "The data sink invocation that should be configured with the template contents",
-          content = @Content(schema = @Schema(implementation = DataSinkInvocation.class)))
-      @org.springframework.web.bind.annotation.RequestBody DataSinkInvocation invocation) {
+                   content = @Content(schema = @Schema(implementation = DataSinkInvocation.class)))
+      @org.springframework.web.bind.annotation.RequestBody DataSinkInvocation invocation
```
bossenti, line 178: same as above

streampipes-platform-services/src/main/java/org/apache/streampipes/ps/PipelineElementTemplateResource.java
```diff
@@ -196,26 +204,27 @@ public ResponseEntity<DataProcessorInvocation> getPipelineElementForTemplate(
       @RequestParam("overwriteNames") String overwriteNameAndDescription,
 
       @RequestBody(description = "The data processor invocation that should be configured with the template contents",
-          content = @Content(schema = @Schema(implementation = DataProcessorInvocation.class)))
-      @org.springframework.web.bind.annotation.RequestBody DataProcessorInvocation invocation) {
+                   content = @Content(schema = @Schema(implementation = DataProcessorInvocation.class)))
+      @org.springframework.web.bind.annotation.RequestBody DataProcessorInvocation invocation
```
bossenti, line 208: same as above

streampipes-platform-services/src/main/java/org/apache/streampipes/ps/PipelineElementTemplateResource.java
```diff
@@ -225,8 +234,9 @@ public ResponseEntity<AdapterDescription> getPipelineElementForTemplate(
       @RequestParam("overwriteNames") String overwriteNameAndDescription,
 
       @RequestBody(description = "The adapter that should be configured with the template contents",
-          content = @Content(schema = @Schema(implementation = AdapterDescription.class)))
-      @org.springframework.web.bind.annotation.RequestBody AdapterDescription adapterDescription) {
+                   content = @Content(schema = @Schema(implementation = AdapterDescription.class)))
+      @org.springframework.web.bind.annotation.RequestBody AdapterDescription adapterDescription
```
bossenti, line 238: same as above

streampipes-platform-services/src/main/java/org/apache/streampipes/ps/PipelineElementTemplateResource.java
bossenti: Ah okay, didn't see this
Does it make sense to keep both?

tenthe: I assume both are required. One is for spring's `RequestBody` and the other is for swagger.
Perhaps there is a way to harmonize this, but I would suggest doing this in another PR as it is independent of the changes from this PR.

bossenti (approved):

Issue #2568, opened 2024-03-17
Make data explorer time series axes configurable


dominikriemer:
### Body

In the data explorer, add a configuration option to the time series chart to customize the y axis range.

### StreamPipes Committer

I acknowledge that I am a maintainer/committer of the Apache StreamPipes project.

PR #2569, opened 2024-03-17
feat(#2568): Make time series axes configurable

dominikriemer:
### Purpose

See #2568 

### Remarks

### DIFF
--- a/ui/src/app/data-explorer/components/widgets/time-series-chart/config/time-series-chart-widget-config.component.ts
+++ b/ui/src/app/data-explorer/components/widgets/time-series-chart/config/time-series-chart-widget-config.component.ts
@@ -79,7 +79,7 @@ export class TimeSeriesChartWidgetConfigComponent extends BaseWidgetConfig<
 
         const lenBefore = Object.keys(currentAxis).length;
 
-        numericPlusBooleanFields.map((field, index) => {
+        numericPlusBooleanFields.forEach((field, index) => {
             const name = field.fullDbName + field.sourceIndex;
             if (!(name in currentColors)) {
                 currentColors[name] = this.presetColors[lenBefore + index];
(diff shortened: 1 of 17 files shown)


ui/src/app/data-explorer/models/dataview-dashboard.model.ts
```diff
@@ -104,6 +104,12 @@ export interface FieldProvider {
     nonNumericFields: DataExplorerField[];
 }
 
+export interface AxisConfig {
+    axisScaleMode: boolean;
```
tenthe, line 108: What is the axisScaleMode? what do you think about renaming it to `autoScaleActive`?

ui/src/app/data-explorer/components/widgets/time-series-chart/config/time-series-chart-widget-config.component.ts
```diff
@@ -146,6 +146,16 @@ export class TimeSeriesChartWidgetConfigComponent extends BaseWidgetConfig<
                 },
             );
         config.showSpike ??= true;
+        config.leftAxis ??= {
```
tenthe, line 149: Does this mean that automatic scaling is active and a minimum and a maximum are set? Or am I misunderstanding this? If this is the case, why do both have to be set?

tenthe (commented):
Thanks for the PR, that's definitely a useful feature. 
The new settings can be set for the right and left axis and not for the individual properties, right?

ui/src/app/data-explorer/components/widgets/time-series-chart/config/time-series-chart-widget-config.component.ts
dominikriemer: Yes, it's just the default value in case autoScale is deselected

tenthe (approved):

Issue #2587, opened 2024-03-19
Add data zoom option to time series chart


dominikriemer:
### Body

In the data explorer, add a configuration option to activate the Echarts data zoom feature.

### StreamPipes Committer

I acknowledge that I am a maintainer/committer of the Apache StreamPipes project.

PR #2588, opened 2024-03-19
feat(#2587): Add data zoom option

dominikriemer:
### Purpose

Allow better zooming in time series charts.

image

### Remarks


tenthe:
Hi @dominikriemer,
why is the configuration in the appearance tab? 
Is this option also available for other widgets?

dominikriemer:
@tenthe I put this into the appearance tab so that settings which don't affect the visualization itself can be found here.
E.g., things like:
* Tooltips
* Legends
* Data Zoom controls
* Toolbox
* probably also axis label settings (but not axis scaling, which changes the actual visualization)

Does it make sense?

Currently, the data zoom is only available in the time series chart, but the configuration is handled by its own component so it shouldn't be hard to add it to other charts as well

tenthe (approved):

PR #2604, opened 2024-03-21
fix: Deactivate dependency check plugin

dominikriemer:
### Purpose

Deactivate the Maven dependency check plugin since it causes timeouts when downloading NVD data and lets our builds fail.

### Remarks


bossenti:
see #2267

bossenti (approved):

PR #2612, opened 2024-03-22
fix: Reduce retry interval, reduce logging

dominikriemer:
### Purpose

Reduce the retry interval for service discovery and reduce set adapter monitoring logging to debug to reduce logger noisiness.

### Remarks


tenthe:
Why did you reduce the amount of retries?
Could this be a problem on slow instances during the initial installation?

PR #2616, opened 2024-03-22
refactor(#2615): Change datatypes from class to string

### LINKED ISSUE #2615: Move all types and semantic types definition to one location in ui
### Body

There are different definitions for types and semantic types in the following UI code. The following classes should be harmonized:

- `core-services/semantic-type/semantic-type.service`
- `dashboard/sdk/model/datatypes`
- `connect/services/datatype.service`.

Additionally, the class `platform-services/src/lib/apis/semantic-types-rest.service` provides the functionality for auto-completion of semantic types by calling the backend to provide suggestions.

One open question is where we should place the type definitions. We could move them to the "platform-services" or "core-services".
Any thought on that?

### StreamPipes Committer

tenthe:
### Purpose

Fix #2615.

In this PR I have harmonized all places in the UI that use **data types** or **semantic data types** (domain properties). 
There are now two static classes `DataTypeService` & `SemanticTypeService` in the `platform-services`.

**However, there are still som open points to clarify. See in the comments below.**

### Remarks


tenthe:
Currently, the number data type is handled inconsistently in StreamPipes. Sometimes the **data type** and sometimes the **semantic data type** is checked. My suggestion would be the following:
	- Remove `'http://www.w3.org/2001/XMLSchema#number` from the data type and use only basic data types
	- Move `SemanticTypeService.isNumber` to `DataTypeService` and check the type instead of the semantic type (`domainProperties`)

tenthe:
Should we also create variables for data type names that are displayed to the user in `DataTypeService`? (e.g. String, Integer, Boolean, ...). If so, how should the variables for the urls and the labels be named?

tenthe:
Should we rename `DataTypeService` to `DataType` & `SemanticTypeService` to `SemanticType`? Because it only contains static variables and there is no need to inject the service into components.

ui/projects/streampipes/platform-services/src/lib/model/types/data-type.service.ts
```diff
@@ -0,0 +1,45 @@
...
+    public static readonly XSD: string = 'http://www.w3.org/2001/XMLSchema#';
+
+    public static readonly INTEGER = this.XSD + 'integer';
+    public static readonly LONG = this.XSD + 'long';
+    public static readonly FLOAT = this.XSD + 'float';
+    public static readonly BOOLEAN = this.XSD + 'boolean';
+    public static readonly STRING = this.XSD + 'string';
+    public static readonly DOUBLE = this.XSD + 'double';
+
+    // Should we support data type number?
```
bossenti, line 29: Do we need it for some reason?

In my mind double, integer, long and float are enough data types to distinguish numerical values.

Maybe we can restrict it on a semantic data type (see `SemanticTypeService`)?

ui/projects/streampipes/platform-services/src/lib/model/types/semantic-type.service.ts
```diff
@@ -0,0 +1,59 @@
...
+        return property.domainProperties.includes(
+            SemanticTypeService.TIMESTAMP,
+        );
+    }
+
+    public static isImage(property: EventProperty): boolean {
+        return property.domainProperties.includes(SemanticTypeService.IMAGE);
+    }
+
+    // TODO should we move this to DataTypeService, and check for the data type instead of domain properties?
```
bossenti, line 53: IMO, only if we keep number as a data type

ui/src/app/dashboard/sdk/ep-requirements.ts
```diff
@@ -51,31 +51,35 @@ export class EpRequirements {
...
 
     static imageReq(): EventPropertyUnion {
-        return EpRequirements.domainPropertyReq('https://image.com');
+        return EpRequirements.domainPropertyReq(SemanticTypeService.IMAGE);
     }
 
     static latitudeReq(): EventPropertyUnion {
-        return EpRequirements.domainPropertyReq(Vocabulary.GEO + 'lat');
+        return EpRequirements.domainPropertyReq(
+            'http://www.w3.org/2003/01/geo/wgs84_pos#lat',
```
bossenti, line 59: Don't we want to include them in the `SemanticTypeService`?

bossenti:
> Currently, the number data type is handled inconsistently in StreamPipes. Sometimes the **data type** and sometimes the **semantic data type** is checked. My suggestion would be the following: - Remove `'http://www.w3.org/2001/XMLSchema#number` from the data type and use only basic data types - Move `SemanticTypeService.isNumber` to `DataTypeService` and check the type instead of the semantic type (`domainProperties`)

As already indicated by my review comments, I'm in favor of your suggestion

bossenti:
> Should we also create variables for data type names that are displayed to the user in `DataTypeService`? (e.g. String, Integer, Boolean, ...). If so, how should the variables for the urls and the labels be named?

Do we use them at multiple places? I could imagine to also include them in `DataType` as a `displayLabel` or similar

bossenti:
> Should we rename `DataTypeService` to `DataType` & `SemanticTypeService` to `SemanticType`? Because it only contains static variables and there is no need to inject the service into components.

From my point of view the `Service` as part of class names is slightly misleading, but I'm also not familiar with the naming conventions in Angular & TypeScript.

tenthe:
> > Should we also create variables for data type names that are displayed to the user in `DataTypeService`? (e.g. String, Integer, Boolean, ...). If so, how should the variables for the urls and the labels be named?
> 
> Do we use them at multiple places? I could imagine to also include them in `DataType` as a `displayLabel` or similar

There are currently only two places where we use it. That's why I wasn't sure.

tenthe:
> > > > Should we also create variables for data type names that are displayed to the user in `DataTypeService`? (e.g. String, Integer, Boolean, ...). If so, how should the variables for the urls and the labels be named?
> > > 
> > > 
> > > Do we use them at multiple places? I could imagine to also include them in `DataType` as a `displayLabel` or similar
> > 
> > 
> > There are currently only two places where we use it. That's why I wasn't sure.
> 
> I think then we can move it to the `DataType` class so we have everything in one place

Ok, I checked it again and there is only one place where it is used. So I'll leave it at that for now.

dominikriemer:
> > Currently, the number data type is handled inconsistently in StreamPipes. Sometimes the **data type** and sometimes the **semantic data type** is checked. My suggestion would be the following: - Remove `'http://www.w3.org/2001/XMLSchema#number` from the data type and use only basic data types - Move `SemanticTypeService.isNumber` to `DataTypeService` and check the type instead of the semantic type (`domainProperties`)
> 
> As already indicated by my review comments, I'm in favor of your suggestion

We need to be careful when changing data types since these are also used by the core (e.g., by the matching feature). Some UI features such as the field selection in the data explorer also depend on these XSD defintions which are provided by the backend. 
So I'm afraid we can't simplify these identifiers safely without major migration effort.

I think the code structure of the refactoring looks good apart from that.

Just be careful when importing `platform-services` dependencies. Due to a misconfiguration in the `tsconfig` file, the IDE often imports these as relative dependencies instead of the correct `@streampipes/platform-services` import, which causes component collisions at build time. I have already started to improve the typescript config in the Angular 17 branch.

dominikriemer (approved):

Issue #232, opened 2022-11-26
Required/optional static properties


dominikriemer:
Extend the SDK to allow for the definition of optional static properties (e.g., username/password vs anonymous authentication) that can be provided, but are not mandatory.

Imported from Jira [STREAMPIPES-13](https://issues.apache.org/jira/browse/STREAMPIPES-13). Original Jira may contain additional context.
Reported by: riemer.

PR #2653, opened 2024-03-27
feat(#232): Support optional free text properties

dominikriemer:
### Purpose

Allow to mark free text static properties as `optional` so that fields in the adapter and pipeline element configuration panels can be left empty if the `optional` flag is set.

We had a field `valueRequired` in the model definition which I have removed since it wasn't used anywhere and defaulted to false, although we currently regard all static properties as required.

### Remarks

### DIFF
--- a/streampipes-model/src/main/java/org/apache/streampipes/model/staticproperty/RemoteOneOfStaticProperty.java
+++ b/streampipes-model/src/main/java/org/apache/streampipes/model/staticproperty/RemoteOneOfStaticProperty.java
@@ -52,7 +52,7 @@ public class RemoteOneOfStaticProperty extends StaticProperty {
     this.valueFieldName = valueFieldName;
     this.labelFieldName = labelFieldName;
     this.descriptionFieldName = descriptionFieldName;
-    this.valueRequired = valueRequired;
+    this.optional = valueRequired;
     this.options = new ArrayList<>();
   }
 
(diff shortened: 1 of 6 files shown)


bossenti (approved):

streampipes-model/src/main/java/org/apache/streampipes/model/staticproperty/RemoteOneOfStaticProperty.java
```diff
@@ -52,7 +52,7 @@ public RemoteOneOfStaticProperty(String internalName, String label, String descr
     this.valueFieldName = valueFieldName;
     this.labelFieldName = labelFieldName;
     this.descriptionFieldName = descriptionFieldName;
-    this.valueRequired = valueRequired;
+    this.optional = valueRequired;
```
tenthe, line 55: Does this have the right meaning here? Because if valueRequired=true, then optional should be false, right?
Do we have to rename this variable?

tenthe (changes requested):

streampipes-model/src/main/java/org/apache/streampipes/model/staticproperty/RemoteOneOfStaticProperty.java
dominikriemer: @tenthe `RemoteOneOfStaticProperty` isn't used anymore and deprecated since 2019, I propose to open a separate PR after merging this one to delete the property completely. The change here is just due to automatic refactoring.

tenthe (approved):

Issue #2659, opened 2024-03-28
Show adapter status light in adapter overview


dominikriemer:
### Body

Similar to pipelines, show a status light indicating if the adapter has been started in the adapter overview.
Health info can be added to the status light later on.

### StreamPipes Committer

I acknowledge that I am a maintainer/committer of the Apache StreamPipes project.

PR #2660, opened 2024-03-28
feat(#2659): Show status light in adapter overview

dominikriemer:
### Purpose

See #2659 

### Remarks


tenthe (commented):
To structure the component a little better, I would suggest making the status light a separate component

bossenti (approved):

PR #2665, opened 2024-03-28
refactor: convert model classes to records

bossenti:
### Purpose

Clean up some model classes by switching from classes to records.

### Remarks

### DIFF
--- a/streampipes-user-management/src/main/java/org/apache/streampipes/user/management/service/TokenService.java
+++ b/streampipes-user-management/src/main/java/org/apache/streampipes/user/management/service/TokenService.java
@@ -28,9 +28,8 @@ public class TokenService {
   public RawUserApiToken createAndStoreNewToken(String email,
                                                 RawUserApiToken baseToken) {
     UserAccount user = getUserStorage().getUserAccount(email);
-    RawUserApiToken generatedToken = TokenUtil.createToken(baseToken.getTokenName());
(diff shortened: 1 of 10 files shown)


streampipes-user-management/src/main/java/org/apache/streampipes/user/management/service/TokenService.java
```diff
@@ -28,9 +28,8 @@ public class TokenService {
   public RawUserApiToken createAndStoreNewToken(String email,
                                                 RawUserApiToken baseToken) {
     UserAccount user = getUserStorage().getUserAccount(email);
-    RawUserApiToken generatedToken = TokenUtil.createToken(baseToken.getTokenName());
+    RawUserApiToken generatedToken = TokenUtil.createToken(baseToken.tokenName());
     storeToken(user, generatedToken);
-    generatedToken.setHashedToken("");
```
bossenti, line 33: Did we do this for a certain reason? From my point of view, it just does not make sense

dominikriemer: I think we just don't want to expose the hashed token to the user.

bossenti: Does the RawUserApiToken then even need this attribute? The only usage I can find is in a test

bossenti:
WIP

PR #2714, opened 2024-04-08
Fix：CheckCompletedVisitor outOfBounds

yanfeng1992:
### Purpose

Using mqtt Adapter, when there is a timestamp field in the data, the original timestamp field is not deleted. New timestamp added. Then checked the Persist events option, and an error occurred when saving.

0880c4c80710b99ea319

0880c4c80710b9e0e41d

image

### Remarks

no

no


yanfeng1992:
Do you think it needs to be added to the 0.95 milestone?  I think this is an issue that needs to be fixed @dominikriemer

bossenti:
@yanfeng1992 thank you very much for your contribution 🙏🏼 
Yes, this should (and will) of course be included in our next release.

bossenti (approved):

tenthe (approved):

PR #2757, opened 2024-04-23
refactor: remove direct usage of couchDB client

bossenti:
### Remarks

### DIFF
--- a/streampipes-data-explorer/src/main/java/org/apache/streampipes/dataexplorer/DataExplorerSchemaManagement.java
+++ b/streampipes-data-explorer/src/main/java/org/apache/streampipes/dataexplorer/DataExplorerSchemaManagement.java
@@ -24,14 +24,7 @@ import org.apache.streampipes.model.datalake.DataLakeMeasure;
 import org.apache.streampipes.model.datalake.DataLakeMeasureSchemaUpdateStrategy;
 import org.apache.streampipes.model.schema.EventProperty;
 import org.apache.streampipes.storage.api.IDataLakeStorage;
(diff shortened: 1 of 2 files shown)


streampipes-data-explorer/src/main/java/org/apache/streampipes/dataexplorer/DataExplorerSchemaManagement.java
```diff
@@ -126,34 +117,20 @@ public void deleteMeasurement(String elementId) {
...
-                                         .query(JsonObject.class);
-
-    for (JsonObject document : docs) {
-      if (document.get("measureName")
-                  .toString()
+    List<DataLakeMeasure> docs = dataLakeStorage.getAllDataLakeMeasures();
+
+    for (var measure : docs) {
+      if (measure.getMeasureName()
                   .replace("\"", "")
```
bossenti, line 124: I'd assume that I can replace the two `replace()` statements as well. What do you think? Do I miss anything?

tenthe (approved):

streampipes-data-explorer/src/main/java/org/apache/streampipes/dataexplorer/DataExplorerSchemaManagement.java
tenthe: Sorry, I am also not quite sure why they are there.

bossenti: Performed some investigation: The replace statements are not necessary (anymore).

PR #2803, opened 2024-04-30
refactor: abstract data explorer modules from Influx-related code

bossenti:
### Purpose

This PR continues the work of #2795 and abstracts the data explorer code from Influx-specific implementations. The idea is that whenever we want to interact with our data explorer storage we do not point to storage-specific implementations but use the `DataExplorerDispatcher` instead. The `DataExplorerDispatcher` gives aces to an an instance of `DataExplorerManager` which is tied to the underlying storage dependent on the configuration of StreamPipes. Therefore, this PR introduces the environment variable `SP_TS_STORAGE` which determines the storage to be used and defaults to `influxDB`.
As for now, Influx is the only storage supported but this allows us to better separate our codebase and to integrate other time series storages as well.

### Remarks

#### Breaking Changes
This PR strongly moves files between different modules and therefore import paths might change, especially with respect to the module `streampipes-data-explorer-influx`. This PR does not have any affect on the behavior or any feature of StreamPipes.


streampipes-commons/src/main/java/org/apache/streampipes/commons/constants/Envs.java
```diff
@@ -54,6 +54,7 @@ public enum Envs {
 
 
   // Time Series Storage
+  SP_TS_STORAGE("SP_TS_STORAGE", "influxDB"),
```
tenthe, line 57:
```suggestion
  SP_TS_STORAGE("SP_TS_STORAGE", "influxdb"),
```

streampipes-data-explorer-management/src/main/java/org/apache/streampipes/dataexplorer/management/DataExplorerDispatcher.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ */
+
+package org.apache.streampipes.dataexplorer.management;
+
+import org.apache.streampipes.commons.environment.Environments;
+import org.apache.streampipes.dataexplorer.api.IDataExplorerManager;
+import org.apache.streampipes.dataexplorer.influx.DataExplorerManagerInflux;
+
+public enum DataExplorerDispatcher {
```
tenthe, line 25: Does it make sense to change this to a class that we can more easily mock for testing instead of a singleton?

streampipes-data-explorer-management/src/main/java/org/apache/streampipes/dataexplorer/management/SupportedDataExplorerStorages.java
```diff
@@ -16,10 +16,16 @@
...
+ * <p>
+ * Supported time series storage implementations:
+ * <ul>
+ *     <li>{@link #INFLUX_DB}: Represents the InfluxDB time series storage implementation.</li>
+ * </ul>
  */
-package org.apache.streampipes.dataexplorer.influx.migrate;
\ No newline at end of file
+public class SupportedDataExplorerStorages {
+  public static final String INFLUX_DB = "influxDB";
```
tenthe, line 30: influxdb?

tenthe (commented):
Hi @bossenti,

thanks for the PR.
I like the idea of the env variable. 

I just have some minor questions, see comments.

Cheers,
Philipp

streampipes-data-explorer-management/src/main/java/org/apache/streampipes/dataexplorer/management/DataExplorerDispatcher.java
bossenti: I have changed `DataExplorerDispatcher` to a regular class, check if that suits your expectations

tenthe (approved):

PR #2841, opened 2024-05-08
guessSchema the encrypted secretValue should be decrypted when used

yanfeng1992:
### Purpose

The encrypted secretValue should be decrypted when used

### Remarks

no
no

### DIFF
--- a/streampipes-connect-management/src/main/java/org/apache/streampipes/connect/management/management/GuessManagement.java
+++ b/streampipes-connect-management/src/main/java/org/apache/streampipes/connect/management/management/GuessManagement.java
@@ -31,6 +31,7 @@ import org.apache.streampipes.model.connect.guess.AdapterEventPreview;
 import org.apache.streampipes.model.connect.guess.GuessSchema;
 import org.apache.streampipes.model.extensions.svcdiscovery.SpServiceTag;
 import org.apache.streampipes.model.monitoring.SpLogMessage;
+import org.apache.streampipes.resource.management.secret.SecretProvider;
 import org.apache.streampipes.serializers.json.JacksonSerializer;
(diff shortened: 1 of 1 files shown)


dominikriemer:
Hi @yanfeng1992 did you have any problem with the current setup?
Secrets should be decrypted by the core before they are sent to the extension service (to avoid that the extensions need to be aware of the encryption passcode) so that no decryption should be necessary when extracting a secret value.

yanfeng1992:
I discovered that the current secret was not decrypted by the core before being sent to the extension service, and the extension service still received the encrypted password. @dominikriemer

yanfeng1992:
I discovered it when adapting and modifying the configuration. Can it be fixed in core? I haven't found a way to fix it in core. 
Is it possible to add the decryption fix directly to the core?

dominikriemer:
Hi @yanfeng1992 thanks for asking!
For pipeline elements, decryption is handled in `streampipes-pipeline-management` in the `SecretEncryptionTask`, which calls a `SecretService` that loops over all configuration and decrypts secrets. For adapters, this class is used within the `WorkerRestClient` of `streampipes-connect-management`. In which cases did you experience the problem? Does this occur always or only for nested structures or something like that?

yanfeng1992:
> Hi @yanfeng1992 thanks for asking! For pipeline elements, decryption is handled in `streampipes-pipeline-management` in the `SecretEncryptionTask`, which calls a `SecretService` that loops over all configuration and decrypts secrets. For adapters, this class is used within the `WorkerRestClient` of `streampipes-connect-management`. In which cases did you experience the problem? Does this occur always or only for nested structures or something like that?

Thanks for your answer, I found the cause of the problem. It is in the guessSchema function that the secretValue is not decrypted. We can fix this problem in core. Please review my latest commit @dominikriemer

bossenti (approved):

PR #2864, opened 2024-05-15
fix: remove gen_ref_pages page from python client docs

ittuann:
### Purpose

The `docs/scripts/gen_ref_pages.py` script should not be directly configured in the docs directory. Because MKDocs creates web pages for all content in the docs directory by default, and the mkdocs-jupyter plugin recognizes and converts it into a notebook page.
This results in the gen_ref_pages.py script content being inappropriately displayed in the official documentation.

We can currently view the script converted into a web page by manually accessing the following link:
<https://streampipes.apache.org/docs/docs/python/latest/scripts/gen_ref_pages/>
Additionally, this page is not indexed in the navigation bar and can only be accessed by manually entering the link.

I believe displaying this script in the official documentation is not the intended outcome. After PR fixes the build results will not contain the contents of the scripts directory.

### Remarks

no
no

### DIFF
--- a/streampipes-client-python/mkdocs.yml
+++ b/streampipes-client-python/mkdocs.yml
@@ -61,6 +61,9 @@ extra:
   version:
     provider: mike
 
+exclude_docs: |
+  /scripts/**
+
 # Extensions
 markdown_extensions:
   - admonition


bossenti (approved):
Thank you very much for housekeeping our python documentation @ittuann 🙏🏼 

Please feel free to address any other issue to observe, if you have any ideas for improvements I'm happy to discuss them as well 🙂

PR #2871, opened 2024-05-16
fix: Replace hard-coded identifier in VersionEndpoint.get() method

ittuann:
### Purpose

The `get` method in the `VersionEndpoint` class internally hard-codes the `identifier` parameter to an empty string `""`, which can be misleading.

When users call the `get` method and pass any identifier, the method always uses an empty string as the identifier. This may lead users to mistakenly believe that their incorrect get request has successfully returned the correct result.

image

With this fix, when users provide an invalid identifier, the method will now throw an appropriate error, ensuring that users can promptly detect the issue and make corrections.

image

Additionally, I also added tests to ensure that the new error handling behaves as expected.

### Remarks

no
no

### DIFF
--- a/streampipes-client-python/streampipes/endpoint/api/version.py
+++ b/streampipes-client-python/streampipes/endpoint/api/version.py
@@ -121,7 +121,8 @@ class VersionEndpoint(APIEndpoint):
         """
         if identifier:
             raise ValueError(
-                "Non-empty 'identifier' is not supported by this endpoint. Please set 'identifier' to an empty string or 'None'."
+                "Non-empty 'identifier' is not supported by this endpoint. "
+                "Please set 'identifier' to an empty string or 'None'."
             )
 
         return super().get(identifier="")
--- a/streampipes-client-python/streampipes/endpoint/api/version.py
+++ b/streampipes-client-python/streampipes/endpoint/api/version.py
@@ -111,7 +111,7 @@ class VersionEndpoint(APIEndpoint):
 
         Raises
         ------
-        NotImplementedError
+        ValueError
             Non-empty 'identifier' is not supported by this endpoint. Please set 'identifier' to an empty string.
 
         Returns
--- a/streampipes-client-python/streampipes/endpoint/api/version.py
+++ b/streampipes-client-python/streampipes/endpoint/api/version.py
@@ -112,7 +112,7 @@ class VersionEndpoint(APIEndpoint):
         Raises
         ------
         ValueError
-            Non-empty 'identifier' is not supported by this endpoint. Please set 'identifier' to an empty string.
+            Non-empty `identifier` is not supported by this endpoint. Please set `identifier` to an empty string or `None`.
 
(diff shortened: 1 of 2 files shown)


streampipes-client-python/streampipes/endpoint/api/version.py
```diff
@@ -109,11 +109,20 @@ def get(self, identifier: str, **kwargs) -> Resource:
...
+        ------
+        NotImplementedError
+            Non-empty 'identifier' is not supported by this endpoint. Please set 'identifier' to an empty string.
+
         Returns
         -------
         versions: Version
             The specified resource as an instance of the corresponding model class([Version][streampipes.model.resource.Version]).  # noqa: 501
         """
+        if identifier != "":
```
bossenti, line 122: Shall we change that to `if identifier:`? This would allow the user to pass `None` as well which is semantically correct.

streampipes-client-python/streampipes/endpoint/api/version.py
```diff
@@ -109,11 +109,20 @@ def get(self, identifier: str, **kwargs) -> Resource:
...
+            raise NotImplementedError(
```
bossenti, line 123: I suggest to use `ValueError`
This is not the intended use for `NotImplementedError` (https://docs.python.org/3/library/exceptions.html#NotImplementedError).

ittuann: Thank you for your feedback! I understand your point and will modify the code as per your suggestion.

streampipes-client-python/streampipes/endpoint/api/version.py
```diff
@@ -109,11 +109,20 @@ def get(self, identifier: str, **kwargs) -> Resource:
         identifier: str
             Not supported by this endpoint, is set to an empty string.
 
+        Raises
+        ------
+        ValueError
+            Non-empty 'identifier' is not supported by this endpoint. Please set 'identifier' to an empty string.
```
bossenti, line 115:
```suggestion
            Non-empty `identifier` is not supported by this endpoint. Please set `identifier` to an empty string or `None`.
```

streampipes-client-python/streampipes/endpoint/api/version.py
```diff
@@ -109,11 +109,20 @@ def get(self, identifier: str, **kwargs) -> Resource:
...
+
         Returns
         -------
         versions: Version
             The specified resource as an instance of the corresponding model class([Version][streampipes.model.resource.Version]).  # noqa: 501
         """
+        if identifier:
+            raise ValueError(
+                "Non-empty 'identifier' is not supported by this endpoint. Please set 'identifier' to an empty string."
```
bossenti, line 124:
```suggestion
                "Non-empty 'identifier' is not supported by this endpoint. Please set 'identifier' to an empty string or 'None'."
```

bossenti (approved):

PR #2873, opened 2024-05-16
#2661 Added error message when self registration fails

### LINKED ISSUE #2661: No user feedback when self registration fails

UI

### What happened?

If the self-registration of a new user fails, e.g, due to misconfigured Email settings, the UI does not display any feedback, whereas it does when the user already exists.
email-registration

### How to reproduce?

_No response_

### Expected behavior

Show an error message in the UI for internal errors (response code 500) as well

_No response_

No, I don't want to submit a PR.

IsaakKrut:
### Purpose
fixes #2661. 
UI shows "Internal Server Error" when 500 is returned

image

### Remarks 


bossenti:
Can we put a more user-friendly notification text here? E.g., something like: "User registration failed - please report this to your admin"

IsaakKrut:
I updated the error message and moved it to the backend. Please let me know if there is a better way to handle the exception. In the commit before the last one, I added a new exception but don't think that was necessary
image

dominikriemer (approved):

PR #2883, opened 2024-05-21
fix: Fix python client Flake8 errors

ittuann:
### Purpose

```
tests/client/test_client.py:141:121: E501 line too long (134 > 120 characters)
```

This issue also causes pre-commit checks to fail.

### Remarks

PR introduces (a) breaking change(s): <yes/no>
no
PR introduces (a) deprecation(s): <yes/no>
no

PR #2884, opened 2024-05-21
fix: Resolve python client mypy failed in pre-commit

ittuann:
### Purpose

```
mypy.....................................................................Failed
- hook id: mypy
- duration: 5.81s
- exit code: 1

Success: no issues found in 5 source files
Success: no issues found in 5 source files
Success: no issues found in 5 source files
Success: no issues found in 5 source files
Success: no issues found in 5 source files
Success: no issues found in 5 source files
Success: no issues found in 5 source files
tests\client\test_versions.py:47: error: "Resource" has no attribute
"backend_version"  [attr-defined]
            self.assertEqual("development", result.backend_version)
                                            ^~~~~~~~~~~~~~~~~~~~~~
Found 1 error in 1 file (checked 5 source files)
Success: no issues found in 5 source files
Success: no issues found in 5 source files
Success: no issues found in 5 source files
Success: no issues found in 5 source files
Success: no issues found in 5 source files
Success: no issues found in 4 source files
```

I analyzed that the issue is due to the `VersionEndpoint` class inheriting from `APIEndpoint` and calling the parent `APIEndpoint` class's `get()` method through `super()` in its `get()` method. Inside the parent `APIEndpoint` class's `get()` method, `response = self._make_request(...)` is used, where the `_make_request()` method returns an instance of the `Response` class. This is what caused mypy to throw an error during static analysis.

However, in actual execution, the object type returned by the `get()` method depends on the class defined by the `_resource_cls()` method implemented by each endpoint (for example, `VersionEndpoint` returns an instance of the `Version` class through `Versions._container_cls`).

Following this, I have updated the relevant code to ensure type consistency, resolving the error found during mypy static analysis.

### Remarks

no
no

### DIFF
--- a/streampipes-client-python/Makefile
+++ b/streampipes-client-python/Makefile
@@ -47,7 +47,7 @@ reformat-all:
 
 .PHONY: pre-commit
 pre-commit:
-	poetry run pre-commit run --verbose --all-files
+	poetry run pre-commit run --verbose --files $(git ls-files 'streampipes/**/*' 'tests/**/*')
 
 .PHONY: unit-tests
 unit-tests:
--- a/streampipes-client-python/Makefile
+++ b/streampipes-client-python/Makefile
@@ -47,7 +47,12 @@ reformat-all:
 
 .PHONY: pre-commit
 pre-commit:
-	poetry run pre-commit run --verbose --files $(git ls-files 'streampipes/**/*' 'tests/**/*')
+	@CHANGED_FILES_PYTHON=$$(git diff --name-only HEAD~1 HEAD -- 'streampipes/**/*' 'tests/**/*'); \
+	if [ -n "$$CHANGED_FILES_PYTHON" ]; then \
+		poetry run pre-commit run --verbose --files $$CHANGED_FILES_PYTHON; \
+	else \
+		echo "No files changed in 'streampipes' or 'tests' directories."; \
+	fi
 
 .PHONY: unit-tests
 unit-tests:
--- a/streampipes-client-python/Makefile
+++ b/streampipes-client-python/Makefile
@@ -47,7 +47,7 @@ reformat-all:
 
 .PHONY: pre-commit
 pre-commit:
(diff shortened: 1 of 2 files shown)


bossenti:
How did you execute pre-commit?
If I'm running `make pre-commit` this does not fail which is due to the fact that it does not watch tests files
We should adapt that as well because such issues should bring the CI to fail

ittuann:
> > CI does not execute tests in the tests directory because the path pattern 'streampipes/**/*' used in the git ls-files command excludes the tests directory.
> 
> Yes I know, so let's fix that 🙂

I would love to contribute together! Please let me know if there are any suggested changes~

ittuann:
The CI failed flake8 results in the CI can be resolved in another PR #2883.

However, the CI results after the changes are different from my local ones.

This could be due to a configuration `working-directory` in GitHub Actions? I saw many files that do not belong to the `streampipes-client-python` directory in the CI console output.
I used the `cd` command locally, and the result was:

```
check json...........................................(no files to check)Skipped
- hook id: check-json
check yaml...........................................(no files to check)Skipped
- hook id: check-yaml
fix end of files.........................................................Passed
- hook id: end-of-file-fixer
- duration: 0.45s
trim trailing whitespace.................................................Passed
- hook id: trailing-whitespace
- duration: 0.42s
```

streampipes-client-python/Makefile
```diff
@@ -47,7 +47,7 @@ reformat-all:
 
 .PHONY: pre-commit
 pre-commit:
-	git ls-files -- 'streampipes/**/*' | xargs poetry run pre-commit run --verbose --files
```
bossenti, line 50: We can do it this way, but since this is used as pre-commit hook I'd prefer to only scan the changed files
Scanning all files causes some unnecessary latency

ittuann: > We can do it this way, but since this is used as pre-commit hook I'd prefer to only scan the changed files Scanning all files causes some unnecessary latency

I got it. I tried replacing `git ls-files` with `git diff` in the latest commit.

PR #2921, opened 2024-06-09
feat: Add "Last Updated Time" for python client docs site

ittuann:
### Purpose

The main site's docs show Last updated time. Python client docs site is now able to do the same after the introduction of this feature.

image

image

### Remarks

no
no


streampipes-client-python/pyproject.toml
```diff
@@ -60,6 +60,7 @@ pandas = ">=1.5, <3"
 pydantic = "~2, >=2.6.3"
 requests = "~2, >=2.28"
 typing-extensions = "~4, >=4.5"
+mkdocs-git-revision-date-localized-plugin = "^1.2.6"
```
bossenti, line 63: Can we please add it to the docs group and pin the version?

bossenti (changes requested):

streampipes-client-python/pyproject.toml
ittuann: Okay I fixed that.

bossenti (approved):

Issue #2729, opened 2024-04-12
Split Array writing to the data lake does not take effect


yanfeng1992:
### Apache StreamPipes version

0.92.0

### Affected StreamPipes components

Processing Elements

### What happened?

![image](https://github.com/apache/streampipes/assets/89568107/f6df9a6b-24b0-40a6-93a4-7b0158f182b1)

There are no fields in the array in influxdb

![image](https://github.com/apache/streampipes/assets/89568107/bde01bca-9ca8-401a-81c1-f8c9059b4bcf)

### Are you willing to submit a PR?

None

PR #2948, opened 2024-06-19
fix(#2729) improve split array processor support nested field

yanfeng1992:
### Purpose

[#2729](https://github.com/apache/streampipes/issues/2729)
improve split array processor support nested field

### Remarks


dominikriemer (commented):
Hi @yanfeng1992 thanks for the PR!
Since the changes also affect the description, we should also add a migration to make sure existing pipelines using the old model are properly updated. 
You can take a look at the MsTeamsSinkMigration as an example and I'm happy to help if you have any questions.

PR #2952, opened 2024-06-22
Fix: Add python client QueryResult model field definitions

ittuann:
### Purpose

Add python client QueryResult model field definitions.

Also I fixed the docstring of `DataLakeMeasureEndpoint.get()`, in the current version its return type is not `DataLakeMeasures` in the docstring but `<class 'streampipes.model.resource.query_ result.QueryResult'>`

### Remarks

no
no

### DIFF
--- a/streampipes-client-python/streampipes/model/resource/query_result.py
+++ b/streampipes-client-python/streampipes/model/resource/query_result.py
@@ -16,7 +16,7 @@
 #
 
 from itertools import chain
-from typing import Any, Dict, List, Literal, Union
+from typing import Any, Dict, List, Literal, Optional, Union
 
 import pandas as pd
 from pydantic.v1 import Field, StrictInt, StrictStr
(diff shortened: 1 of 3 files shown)


streampipes-client-python/streampipes/model/resource/query_result.py
```diff
@@ -71,8 +71,11 @@ def convert_to_pandas_representation(self) -> Dict[str, Union[List[str], List[Li
 
     total: StrictInt
     headers: List[StrictStr]
-    all_data_series: List[DataSeries]
+    all_data_series: List[DataSeries] = Field(alias="allDataSeries")
```
bossenti, line 74: Alias is not needed here (same for the other three fields) 
We have default alias generator in place, that creates the camel case name as alias for every field

bossenti (changes requested):

streampipes-client-python/streampipes/model/resource/query_result.py
ittuann: Ok I removed the added alias.

bossenti (approved):

PR #2968, opened 2024-06-27
chore: Cleanup persistence layer

dominikriemer:
### Purpose

Do some more cleanup to reduce persistence layer complexity.

### Remarks


streampipes-model-client/src/main/java/org/apache/streampipes/model/client/user/UserActivationToken.java
tenthe, line 1: Isn't that something that can be solved with a JSON annotation?

tenthe (approved):

PR #2999, opened 2024-07-05
chore: cli installer switch to compose v2 api

uliludmann:
### Purpose
Hey thanks for providing this awesome piece of software.
While installing with the `cli-installer` I stumbled upon a legacy usage of `docker-compose` which broke on my system. 
https://docs.docker.com/compose/migrate/#what-are-the-differences-between-compose-v1-and-compose-v2

This tiny pr fixes the issue.

### DIFF
--- a/installer/cli/bin/common
+++ b/installer/cli/bin/common
@@ -69,8 +69,8 @@ deployment_notice() {
 
     If your environment contains the StreamPipes UI, open your browser and follow
     the instructions to get started:
-    
-    Local Docker for Mac/Windows, Linux:     http://localhost/   
+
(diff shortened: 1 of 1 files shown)


tenthe:
@uliludmann, thank you for this PR, your contribution is much appreciated and we will merge it soon.

I noticed that we use version `3.4` in the docker-compose.yml files in the installer. Should we also update this? This is already a bit older. Which version would you suggest and do you know if this has any effect?

uliludmann:
Hi, even better! I could refactor the docker compose deployment and even contribute a helm chart. Should we create tickets for it?

tenthe:
Thank you, that would be awesome. 
Yes, you can create an issue for that.

PR #3029, opened 2024-07-15
feat: Implement visit for mapping property nary in pipeline element templates

tenthe:
### Purpose

This functionality is needed to refactor the unit tests for the `StringToStateProcessor` class. The processor requires a `requiredPropertyWithNaryMapping` input from the user. The unit test framework for processing elements uses the pipeline element template functionality.

Is it possible that this new functionality of this PR will break the templates in the UI? Because we also do not have the original event schema.

### Remarks


tenthe:
It seems that the UI prevents a user from saving this information.
Therefore, we should be able to safely merge this PR.

see screenshot:

grafik

dominikriemer (approved):

PR #3101, opened 2024-08-03
chore: Provide adapter and pipeline preview as stream

dominikriemer:
### Purpose

Replace the polling-based provision of stream and pipeline preview with a streaming approach to improve performance and avoid that data gets lost when connections are opened and closed multiple times.

### Remarks


tenthe (approved):
Just out of curiosity, how is the stream stopped when the preview is closed in the UI?

dominikriemer:
> Just out of curiosity, how is the stream stopped when the preview is closed in the UI?

The streaming output should be closed if the request is canceled. In addition, I added a fallback so that the connection is automatically closed by the server after three minutes.

PR #3137, opened 2024-08-13
Add dialog window for deleting data views/dashboards and test functionality with new E2E-test

Marcelfrueh:
### Purpose
 Add a dialog window when the delete button for a data view or dashboard is clicked. The dialog window demands confirmation from the user for deleting a data view or dashboard. To test the functionality of this dialog window, a new test is introduced: it checks that canceling a deletion action keeps the items intact and confirms successful deletion afterward. Moreover, deleteWidget.ts has been modified accordingly to consider the new dialog window.

### Remarks

### DIFF
--- a/ui/cypress/support/utils/datalake/DataLakeUtils.ts
+++ b/ui/cypress/support/utils/datalake/DataLakeUtils.ts
@@ -184,6 +184,20 @@ export class DataLakeUtils {
         cy.dataCy('save-dashboard-btn', { timeout: 10000 }).click();
     }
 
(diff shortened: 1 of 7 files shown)


tenthe (changes requested):
Hi @Marcelfrueh,
Thanks for the test. That looks good.
Would you mind creating utils functions in the `DataLakeUtils` for the different actions?

I'm not quite sure yet what the best API is for our testing framework, but I would suggest that if we have actions that are used multiple times, we put them in the utils. What do you think?

tenthe (approved):

PR #3148, opened 2024-08-16
feat: add semantic type to create_data_stream

raphaelbalzer:
### Purpose
Enable specifying a semantic type for the attributes of a datastream. For example, images can now be included in an output stream.

### Remarks
Instead of a string representing the runtime type, an instance of a dataclass containing information about the semantic type and runtime type needs to be passed as a value for the attribute key.

### DIFF
--- a/streampipes-client-python/streampipes/functions/utils/data_stream_generator.py
+++ b/streampipes-client-python/streampipes/functions/utils/data_stream_generator.py
@@ -17,7 +17,9 @@
 
 from dataclasses import dataclass
 from enum import Enum
-from typing import Dict
+from typing import Dict, Optional
+
+from typing_extensions import deprecated
 
 from streampipes.functions.broker import SupportedBroker
 from streampipes.model.common import (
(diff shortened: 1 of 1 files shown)


dominikriemer (commented):
@raphaelbalzer thanks for the PR! Is it possible to make this backwards compatible by introducing a new method and deprecating the existing method?

streampipes-client-python/streampipes/functions/utils/data_stream_generator.py
```diff
@@ -48,11 +51,26 @@ class RuntimeType(Enum):
     INTEGER = "integer"
     LONG = "long"
 
+class SemanticType(Enum):
```
bossenti, line 54: What is `SemanticType` used for?
I can't find any usage

streampipes-client-python/streampipes/functions/utils/data_stream_generator.py
```diff
@@ -117,3 +136,35 @@ def create_data_stream(
 
     data_stream.element_id = sanitized_stream_id
     return data_stream
+
+# TODO Use an more general approach to create a data stream
+@deprecated('deprecated; please use "compose_data_stream" instead.')
```
bossenti, line 141: Could you please adapt the tests that use `create_data_stream` as well?
In addition, it would be great to have some dedicated tests for the new `compose_data_stream` function, if you are open to it

PR #3168, opened 2024-08-20
#3042 python client updating data stream

### LINKED ISSUE #3042: Automatic Handling of Data Stream Changes in Python Client
## Description
Currently, new data streams can be created in Python and registered in the backend. However, problems arise if the data streams change after a function restart, such as changes to the data stream name or properties in the event schema.

## Expected Behavior
- The Python client should automatically handle changes to data streams if possible.
- If automatic handling is not possible, a warning should be displayed to the user.

## Proposed Solution
- Implement functionality similar to migration when changing adapters.
- Add necessary functions to the REST API, such as updating data streams (PUT).

### StreamPipes Committer

IsaakKrut:
### Purpose
Closes #3042 

### Remarks
When testing I noticed the Data Stream is being updated in the database, but not in the UI. Should I use a different method to update the Data Streams?
image
image

Also, should any other functionality be implemented to automatically handle the changes?

PR introduces (a) deprecation(s): <no

### DIFF
--- a/streampipes-rest/src/main/java/org/apache/streampipes/rest/impl/pe/DataStreamResource.java
+++ b/streampipes-rest/src/main/java/org/apache/streampipes/rest/impl/pe/DataStreamResource.java
@@ -34,6 +34,7 @@ import org.springframework.web.bind.annotation.DeleteMapping;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.PutMapping;
 import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RestController;
@@ -83,6 +84,20 @@ public class DataStreamResource extends AbstractAuthGuardedRestResource {
   public ResponseEntity<?> addDataStream(@RequestBody SpDataStream dataStream) {
     try {
       getDataStreamResourceManager().add(dataStream, getAuthenticatedUserSid());
+      return ok();
+    } catch (IllegalArgumentException e) {
+      return badRequest(e.getMessage());
+    }
+  }
+
+  @PutMapping(
+      produces = MediaType.APPLICATION_JSON_VALUE,
+      consumes = MediaType.APPLICATION_JSON_VALUE
+  )
+  @PreAuthorize(AuthConstants.HAS_WRITE_PIPELINE_ELEMENT_PRIVILEGE)
+  public ResponseEntity<?> updateDataStream(@RequestBody SpDataStream dataStream) {
+    try {
+      getDataStreamResourceManager().update(dataStream);
       return ok();
     } catch (IllegalArgumentException e) {
       return badRequest(e.getMessage());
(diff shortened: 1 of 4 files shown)


tenthe:
@IsaakKrut  thank you for this PR. The code looks good, but before we proceed,  I think we should clarify some conceptual ideas in StreamPipes that are not quite clear yet.

### General Considerations:

- **Adapters vs. Data Streams:** These are two distinct elements in our architecture, but they have interdependencies. We are currently not fully consistent with the concept due to these dependencies.

### Topic 1:

- **Data Source Changes:** Your PR modifies the Data Source only. When we change Data Sources, we may also need to update the Pipelines. The functionality for handling this in Adapters can be found in `AdapterResource.performPipelineMigrationPreflight`. A similar approach might be required for Data Source updates.

### Topic 2:

- **Event Schema Handling:**
    - This is not directly related to your PR, but it is an open issue in our system that your PR brought to light.
    - **Problem:** An Adapter creates a Data Stream. A user then changes the Event Schema of this Data Stream via the API. This can lead to a situation where the Adapter and the Data Source have different Event Schemas.
    - I'm not yet sure how to resolve this. We can discuss this separately if needed.

I think **Topic 2** should be discussed in a separate thread, as there are additional points that need clarification.

How should we proceed with **Topic 1**? We could either:

- **Option 1**: Address it in this PR by implementing the necessary updates for Data Source changes.
- **Option 2**: Leave it for now and create a separate issue and PR after we have a solution for **Topic 2**.

IsaakKrut:
Hi @tenthe, I can work on the option 1. I have a couple of questions though.

1. When updating the data stream should any changes be propagated to the corresponding AdapterDescription? They share fields inherited from NamedStreamPipesEntity. For example, I noticed name and description are set to the same value on adapter creation from the UI.
2. Current updates only update EventSchema on the pipeline DataStream. Should this new API do the same or it should replace the whole DataStream object?

tenthe:
Hi @IsaakKrut,

good questions. Here are some thoughs from my side.

**Propagation changes to AdapterDescription**:  
Propagating changes to the event schema directly to `AdapterDescription` could be problematic, as it might require modifications to the adapter itself, which might be different for each adapter type. Currently we think about that adapters should manage their own event schemas, which would change how we display event streams and adapter instances in the pipeline editor. Then pipelines could either have a data stream or an adapter as a data source. Currently, each adapter creates a data stream as a seperate object that needs to be kept in sync. I suggest we discuss this further in a separate issue.

**Updating DataStream**:  
Regarding this PR I suggest to  propagate event schema changes in the `DataStream` through the pipeline, similar to how we handle adapter updates. Check `AdapterResource.performPipelineMigrationPreflight()` for reference, and the update logic is in `AdapterUpdateManagement.updateAdapter` at line 75.

Feel free to share your thoughts or ask any questions. I'm also not entirely sure what the best solution is, so we can discuss this further.

streampipes-rest/src/main/java/org/apache/streampipes/rest/impl/pe/DataStreamResource.java
```diff
@@ -42,7 +49,17 @@
 
 @RestController
 @RequestMapping("/api/v2/streams")
-public class DataStreamResource extends AbstractAuthGuardedRestResource {
+public class DataStreamResource extends AbstractAdapterResource<AdapterMasterManagement> {
+
+  public DataStreamResource() {
+    super(() -> new AdapterMasterManagement(
```
tenthe, line 55: Can you tell me what the reason is why you added this code? I do not completely understand it.

IsaakKrut: I needed access to AdapterMasterManagement instance. I used the same pattern as in AdapterResource to bring it in

PR #3195, opened 2024-08-28
#2094 adapter migration error handling

### LINKED ISSUE #2094: Handle case for element that is older than all migrations provided
### Body

Within #2001 we introduced a basic support for migrating pipeline elements.
One corner case that was thereby neglected is the following:

Let's assume an existing StreamPipes instance has an adapter of type `AdapterA` that is still in version `0`. This StreamPipes has now updated to a new version that is far ahead of the current one and therefore only provides migrations for `AdapterA` from version `1` onwards.

As of now, this adapter instance is simply not recognized and the adapter is not usable anymore.
This case should also be handled properly, at least by logging it somewhere.

### StreamPipes Committer

IsaakKrut:
### Purpose
Closes #2094

### Remarks
This PR is not ready yet. Looking for help

### DIFF
--- a/streampipes-connect-management/src/main/java/org/apache/streampipes/connect/management/management/AdapterMigrationManager.java
+++ b/streampipes-connect-management/src/main/java/org/apache/streampipes/connect/management/management/AdapterMigrationManager.java
@@ -21,6 +21,7 @@ package org.apache.streampipes.connect.management.management;
 import org.apache.streampipes.commons.exceptions.connect.AdapterException;
 import org.apache.streampipes.manager.migration.AbstractMigrationManager;
 import org.apache.streampipes.manager.migration.IMigrationHandler;
+import org.apache.streampipes.model.connect.adapter.AdapterHealthStatus;
 import org.apache.streampipes.model.extensions.svcdiscovery.SpServiceRegistration;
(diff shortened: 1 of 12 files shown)


IsaakKrut:
Hi @tenthe,

I implemented the front-end changes for adapter notifications. Looking for help with back-end. 

1. I am currently facing an issue when the adapter version is lower than the provided migration's. Health status and notifications are being persisted but the running state is not. I expected WorkerRestClient.stopStreamAdapter to stop the adapter, but I end up with an adapter still in running state, but with the health status and notifications. 
2. Also, please take a look if the back-end logic is updated in the right place. I've updated AdapterMigrationManager.handleMigrations().
3. Let me know what you think about the set of health statuses(same as on the pipeline) and if you think they should be set anywhere else in the code. Maybe something should be added to handle old elements without the AdapterHealthStatus

dominikriemer:
Hi @IsaakKrut great that you are working on this.
I'll need to have a closer look at your first two questions, probably this weekend.
Concerning the health status, it's very cool to have a similar behaviour as pipelines. Do you think there is a problem that occurs for pipeline elements without health status? I'm not 100% sure but when the core is stopped, all adapters are also stopped and I'd expect that the health status is then also set. So the status will be present after an update anyways or am I wrong?

PR #3231, opened 2024-09-16
feat: Manually override dimensions in data lake sink

dominikriemer:
### Purpose

The data lake sink, which stores data to the underlying InfluxDB, uses the `PropertyScope` of the input event to determine which fields are stored as dimension/tag. Users should be able to manually override these settings when using the data lake sink.

The PR adds an additional option to the data lake sink which resolves at runtime and allows users to mark boolean, integer and string fields as dimension.

### Remarks


streampipes-sdk/src/main/java/org/apache/streampipes/sdk/extractor/AbstractParameterExtractor.java
```diff
@@ -666,4 +666,10 @@ private List<EventProperty> getEventPropertiesByScope(PropertyScope scope, Integ
                                                        .equals(scope.name()))
         .collect(Collectors.toList());
   }
+
+  @Override
+  public List<EventProperty> getInputEventProperties(int streamIndex) {
```
tenthe, line 671: The streamIndex is not used here. Is this an error or is it intentional?

streampipes-extensions/streampipes-sinks-internal-jvm/src/main/java/org/apache/streampipes/sinks/internal/jvm/datalake/DataLakeDimensionProvider.java
```diff
@@ -0,0 +1,70 @@
...
+
+import java.net.URI;
+import java.util.List;
+
+public class DataLakeDimensionProvider {
+
+  public void applyOptions(List<EventProperty> inputFields,
+                           RuntimeResolvableAnyStaticProperty staticProperty) {
+    var primitiveFields = getPrimitiveFields(inputFields);
+    primitiveFields
```
tenthe, line 37: I do not completely understand this functionality. Can you please explain why it is first added if not exists and then removed if it is not in the fields array

tenthe (changes requested):

streampipes-extensions/streampipes-sinks-internal-jvm/src/main/java/org/apache/streampipes/sinks/internal/jvm/datalake/DataLakeDimensionProvider.java
dominikriemer: This ensures that if there are updates to the pipeline (e.g., a field is removed), existing selections are kept in case only some fields change. Otherwise, options would be regenerated at every change and existing user selections would be lost.

tenthe (approved):

PR #3251, opened 2024-09-24
refactor: Rename domain properties to semantic type

dominikriemer:
### Purpose

For better clarity, rename `domainProperties` to `semanticType` in the `EventProperty` model. 
Semantic types are now stored as a string instead of a `List<URI>`. 
Old usages of domainProperties should still work and have been marked as deprecated.
We should also change these in the Python and Go client.

### Remarks

* Use `semanticType` instead of `domainProperties` to assign a semantic type to event properties.


streampipes-sdk/src/main/java/org/apache/streampipes/sdk/builder/PrimitivePropertyBuilder.java
```diff
@@ -53,12 +52,12 @@ public static PrimitivePropertyBuilder create(Datatypes datatype, String runtime
   /**
    * Specifies the semantics of the property (e.g., whether a double value stands for a latitude coordinate).
    *
-   * @param domainProperty The domain property as a String. The domain property should reflect an URI. Use some
+   * @param semanticType The semantic type as a String. Use some
    *                       existing vocabulary from {@link org.apache.streampipes.vocabulary} or create your own.
    * @return
    */
```
tenthe, line 58: Add a new method for semanticType and deprecate the `domainProperty` method.

tenthe (changes requested):

tenthe (approved):

PR #3294, opened 2024-10-09
fix: Support dynamic assets in Maven plugin to generate docs

dominikriemer:
### Purpose

Modify the `streampipes-maven-plugin` so that also dynamic assets are resolved when generating the Markdown documentation. Dynamic assets are used by the PLC4X adapters to auto-generate the available configuration options.

### Remarks

### DIFF
--- a/streampipes-extensions-api/src/main/java/org/apache/streampipes/extensions/api/assets/AssetResolver.java
+++ b/streampipes-extensions-api/src/main/java/org/apache/streampipes/extensions/api/assets/AssetResolver.java
@@ -23,8 +23,18 @@ import java.util.Properties;
 
 public interface AssetResolver {
 
-  byte[] getAsset(String assetName) throws IOException;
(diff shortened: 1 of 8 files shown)


tenthe (approved):

tenthe:
**Discussion:** It seems we're currently using the name `Asset` for two different concepts, which might lead to confusion. To make the distinction clearer, I suggest renaming one of them to something more descriptive—perhaps `PipelineElementAsset`. 

What are your thoughts on this proposal? Do you have any alternative name suggestions?

dominikriemer:
> **Discussion:** It seems we're currently using the name `Asset` for two different concepts, which might lead to confusion. To make the distinction clearer, I suggest renaming one of them to something more descriptive—perhaps `PipelineElementAsset`.
> 
> What are your thoughts on this proposal? Do you have any alternative name suggestions?

Yes, good idea, we can harmonize this to `ExtensionAsset` which we already use in the model definition.

PR #3330, opened 2024-11-12
Test job

salmanaslam25:
### Remarks


tenthe:
Thanks for implementing this processor!

I have two minor suggestions for improvement:
- Could we extend it to allow the user to specify how many events should be included in the sum? For example, if the user enters "5," the processor would calculate the sum of the last five events.
- Also, what do you think about renaming the processor to something like SumProcessor or SummationProcessor to better reflect its function?

Let me know your thoughts!

PR #3437, opened 2025-01-23
fix(#3425): Properly store stream transformation rules in adapter

### LINKED ISSUE #3425: Adapter's "Reduce event rate" will not work on edit

UI

### What happened?

When creating an OPCUA adapter (using subscription mode) i entered 10000 in Reduce event rate - i got an changed value as expected every 10s. Show code will NOT display this value.

When edit an existing adapter "reduce event rate" is not checked any longer, "show code" will also NOT display the value entered when creating the adapter. 

When saving (or modifieing) an exiting adapter the value entered in "reduce event rate" will be ignored and deleted! The events will occur at EVERY change!

### How to reproduce?

Create an OPCUA adapter, check "Reduce event rate" enter 10000 and save. After edit and save this value gets lost

### Expected behavior

_No response_

_No response_

None

dominikriemer:
### Purpose

Fixes #3425 

### Remarks


ui/cypress/tests/connect/editAdapterTransformationRulesAreKept.ts
tenthe, line 1: Please rename the file to spec.ts, otherwise it will be ignored by the nightly validation run.

tenthe (changes requested):

tenthe (approved):

PR #3550, opened 2025-03-31
fix: asset filters and labels for assets

Marcelfrueh:
### Purpose
 This fixes the asset filters for the dashboard and data explorer. Moreover, the label selection in asset settings is fixed.

### Remarks

### DIFF
--- a/ui/src/app/assets/components/asset-details/edit-asset/asset-details-panel/asset-details-basics/asset-details-labels/asset-details-labels.component.html
+++ b/ui/src/app/assets/components/asset-details/edit-asset/asset-details-panel/asset-details-basics/asset-details-labels/asset-details-labels.component.html
@@ -63,7 +63,9 @@
                     (optionSelected)="selected($event)"
                 >
                     @for (label of filteredLabels | async; track label) {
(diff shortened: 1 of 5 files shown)


ui/src/app/assets/components/asset-details/edit-asset/asset-details-panel/asset-details-basics/asset-details-labels/asset-details-labels.component.html
```diff
@@ -63,7 +63,9 @@
                     (optionSelected)="selected($event)"
                 >
                     @for (label of filteredLabels | async; track label) {
-                        <mat-option [value]="label">{{ label }}</mat-option>
+                        <mat-option [value]="label.label">{{
```
dominikriemer: Do we need to track by ID here so that it can be determined which label has been selected (if there are two labels having the same name)?

Marcelfrueh: Yes, you're right – thanks! I've updated it to use label._id now.

dominikriemer (approved):

PR #3568, opened 2025-04-20
Add Spotless plugin for automated code formatting

luoluoyuyu:
### Purpose

Add Spotless plugin to realize automatic code formatting. The ```mvn spotless:apply``` command can automatically format the code. Thanks to @Yuukadesu for his contribution. Based on PR https://github.com/apache/streampipes/pull/3290

### Remarks


dominikriemer:
Hi @luoluoyuyu thanks for the PR!  I'm in favor of introducing spotless :-)
Do you think it's possible to turn automated formatting on only for newly modified files? So that we can switch to spotless on a per-module basis instead of having a single huge PR?
And just an idea..what do you think about configuring spotless so that every method parameter is written in a new line? Don't now if this is considered good practice but I personally like it since it's improving the readability of git diffs in code changes. Do you or anyone else have any experience on what is a good choice here?

luoluoyuyu:
> Hi @luoluoyuyu thanks for the PR! I'm in favor of introducing spotless :-) Do you think it's possible to turn automated formatting on only for newly modified files? So that we can switch to spotless on a per-module basis instead of having a single huge PR? And just an idea..what do you think about configuring spotless so that every method parameter is written in a new line? Don't now if this is considered good practice but I personally like it since it's improving the readability of git diffs in code changes. Do you or anyone else have any experience on what is a good choice here?

Great suggestion. I have modified it according to the suggestion. The function parameters are already aligned to columns, and the Spotless plugin is not effective. You need to set Skip to false to make it effective.

dominikriemer (approved):

dominikriemer:
Great, thank you @luoluoyuyu !

PR #3569, opened 2025-04-21
Add LLM Multi-Model Prompt Processor

grainier:
### Purpose

This PR adds a **Multi‑Model Prompt Processor** capable of calling OpenAI, Anthropic and Ollama/Llama models from StreamPipes pipelines.
  * **New processor module:** `streampipes‑processors‑llm‑jvm`
  * Supports stateless, windowed (N‑turn) or full‑history conversations.
  * Exposes a *temperature* slider and validates provider‑specific credentials.
  * Automatically appends the model’s reply as a new event field.

Samples Prompts:
- For a windowed conversation -> `For each input number, calculate and return the current average. Do not output any other information other than the  answer.`
- For a stateless conversation -> `If the input number is greater than 100, return "WARNING". Otherwise return "OK"`.

### Remarks

* New third‑party library: **LangChain4j** – https://github.com/langchain4j/langchain4j (Apache‑2.0). It pulls only lightweight, provider‑specific HTTP clients.

### Questions
1.  Is it acceptable to ship this as a dedicated* **`streampipes‑processors‑llm‑jvm`** *artifact, or should it live under* **`streampipes‑processors‑enricher‑jvm`** *instead?*
2. If a separate artifact is preferred, which distro flavours should include it?*  (i.e. `AllExtensionsInit` (default), `AllExtensionsIIoTInit`, `ExtensionsIIoTMinimalInit`)?


dominikriemer:
HI @grainier very cool!
I think it's good to introduce this as a dedicated artifact. We can add this to `AllExtensionsInit` and `AllExtensionsIIoTInit`.

dominikriemer (approved):

tenthe:
Hi @grainier,

Thanks a lot for your PR! I tested it locally with Ollama, and everything worked perfectly. I really like this new functionality.

I was wondering if we could adjust the processor slightly to support using more than one event property as input. One idea would be to change the current `requiredPropertyWithUnaryMapping` to `requiredPropertyWithNaryMapping`.

For the UserMessage, we could use key-value pairs, where the key is the runtime name and the value is the corresponding property value.

What do you think about this approach?

grainier:
@dominikriemer / @tenthe Thank you for your feedback! I’ve added support for selecting multiple event properties in commit [628a7ae](https://github.com/apache/streampipes/pull/3569/commits/628a7ae959004db1eacf28497630f8eee0d3ae42). Please let me know if you have any suggestions for further improvement.

dominikriemer (approved):

tenthe (approved):

tenthe:
Hi @grainier,
thanks a lot for the changes.
I think this is a really cool processor and I'm excited to start using it.
I would suggest we go ahead and merge it. If someone notices something during use, we can expand the functionality in the future.

We're currently trying to add a bit more content to our blog.
Would you be interested in writing a short blog post about the processor and the kinds of use cases it enables?
We could publish it after the next release.

Cheers,
Philipp

grainier:
Hi @tenthe - Thanks for the feedback. I’d be happy to write a short blog post on the processor and its use cases. I’ll start drafting and let you know when it’s ready.

PR #3570, opened 2025-04-22
Feat: Support TsFileSink

Yuukadesu:
### Remarks


streampipes-extensions/streampipes-sinks-databases-jvm/src/main/java/org/apache/streampipes/sinks/databases/jvm/tsfile/TsFileSink.java
```diff
@@ -0,0 +1,323 @@
...
+import java.util.Random;
+
+public class TsFileSink extends StreamPipesDataSink {
+
+  private static final Logger log = LoggerFactory.getLogger(TsFileSink.class);
+  public static final String DEVICE_ID_KEY = "device_id";
+  public static final String TSFILE_NAME_KEY = "tsfile_name";
+  public static final String TIMESTAMP_MAPPING_KEY = "timestamp_mapping";
+  //To tell where the tsfile is stored.Like "/home/user/".
+  public static final String DIR_ABSOLUTE_PATH_KEY = "dir_absolute_path";
```
luoluoyuyu: I think the name of this configuration should be called ``` TsFile generation directory``` which would be more appropriate

streampipes-extensions/streampipes-sinks-databases-jvm/src/main/java/org/apache/streampipes/sinks/databases/jvm/tsfile/TsFileSink.java
```diff
@@ -0,0 +1,323 @@
...
+    for (Map.Entry<String, Object> measurementValuePair : measurementValuePairs.entrySet()) {
+      if (measurementValuePair.getKey().equals(timestampFieldId)) {
+        continue;
+      }
+      final String measurementId = measurementValuePair.getKey();
+      final Object value = measurementValuePair.getValue();
+
+      if (value instanceof Boolean) {
+        size += BOOLEAN_SIZE;
+        tsRecord.addTuple(DataPoint.getDataPoint(TSDataType.BOOLEAN, measurementId, String.valueOf(value)));
```
luoluoyuyu: Why does it need to be converted into String before writing? There should be other writing methods here?

streampipes-extensions/streampipes-sinks-databases-jvm/src/main/java/org/apache/streampipes/sinks/databases/jvm/tsfile/TsFileSink.java
```diff
@@ -0,0 +1,323 @@
...
+        size += sValue.length();
+        tsRecord.addTuple(DataPoint.getDataPoint(TSDataType.STRING, measurementId, String.valueOf(value)));
+      } else {
+        throw new UnsupportedOperationException("Unsupported data type: " + value.getClass());
+      }
+    }
+
+    try {
+      if (maxTime > timestamp) {
+        log.info("timestamp fault");
```
luoluoyuyu: The log here can be more detailed, for example, it indicates that the file size does not reach the expected size, but because the time of the written measurement point is less than the previously written time, the file needs to be closed in advance, etc.

streampipes-extensions/streampipes-sinks-databases-jvm/src/main/java/org/apache/streampipes/sinks/databases/jvm/tsfile/TsFileSink.java
```diff
@@ -0,0 +1,323 @@
...
+      }
+      if (totalWriteSize >= maxTsFileSize) {
+        resetTsFileWriter();
+        log.info("Success to reset tsFileWriter");
+        return;
+      }
+      if (writeSize >= maxFlushDiskSize) {
+        tsFileWriter.flushAllChunkGroups();
+        writeSize = 0;
+        log.info("Success to flush tsFileWriter,totalWriteSize is " + totalWriteSize);
```
luoluoyuyu: I think this log can be deleted.

streampipes-extensions/streampipes-sinks-databases-jvm/src/main/java/org/apache/streampipes/sinks/databases/jvm/tsfile/TsFileSink.java
```diff
@@ -0,0 +1,323 @@
...
+        totalWriteSize += size;
+        writeSize += size;
+        maxTime = timestamp;
+      } catch (WriteProcessException | IOException e) {
+        resetTsFileWriter();
+        throw new SpRuntimeException("Failed to write TSRecord", e);
+      }
+      if (totalWriteSize >= maxTsFileSize) {
+        resetTsFileWriter();
+        log.info("Success to reset tsFileWriter");
```
luoluoyuyu: I think this log can be deleted.

streampipes-extensions/streampipes-sinks-databases-jvm/src/main/java/org/apache/streampipes/sinks/databases/jvm/tsfile/TsFileSink.java
```diff
@@ -0,0 +1,323 @@
...
+      throw new SpRuntimeException("Failed to create new tsfile");
+    }
+    return tsfile;
+  }
+
+  private void resetTsFileWriter() throws IOException {
+    if (tsFileWriter != null){
+      tsFileWriter.close();
+      log.info("Success to close tsFileWriter,writeSize is " + writeSize + " and totalWriteSize is " + totalWriteSize);
+    }
```
luoluoyuyu:
```suggestion
      log.info("Success to close tsFileWriter, file name is {}, totalWriteSize is {}", newTsFile.getFileName(), totalWriteSize);
    }
```

streampipes-extensions/streampipes-sinks-databases-jvm/src/main/resources/org.apache.streampipes.sinks.databases.jvm.tsfile/documentation.md
```diff
@@ -0,0 +1,56 @@
...
+To create timeStamp field in TsFile.
+
+### DirAbsolutePath
+The absolute path of directory where the TsFile is stored.
+
+### MaxTsFileSize
+The maximum size of the TsFile. The default value is 10MB.
+
+### MaxFlushDiskSize
+The maximum size of the TsFile to flush to disk.
```
luoluoyuyu: I think the description of this configuration item ```MaxFlushDiskSize``` should be the maximum memory cache when writing TSFile

Yuukadesu: Thanks for your suggestions! I will use these suggestions to modify my code.

dominikriemer (approved):

tenthe:
Hi @Yuukadesu,

Thanks a lot for the PR — really appreciate the contribution!

I noticed you're checking the runtime type for each of the values. That makes sense, but I was wondering: could there be any edge cases where the declared type is `float`, but the actual value is something like `1` (which is technically an integer)? I'm not sure if this would cause issues when writing to the TSFile — do you know if it can handle consecutive values with slightly different types like that?

If this turns out to be a potential problem, there's another approach you could consider: checking the event schema directly using the `Event` object:

```java
event.getSchemaInfo().getEventSchema()
```

This gives you a consistent view of the expected types.

Also, if you're looking for type-safe access to the data, you can use:

```java
event.getFieldByRuntimeName("runtimeName").getAsPrimitive().getAsBoolean()
```

Of course, if the current map-based approach works fine with TSFile and handles mixed types well, feel free to stick with it — just wanted to mention this as another option in case it's useful.

Yuukadesu:
> Hi @Yuukadesu,
> 
> Thanks a lot for the PR — really appreciate the contribution!
> 
> I noticed you're checking the runtime type for each of the values. That makes sense, but I was wondering: could there be any edge cases where the declared type is `float`, but the actual value is something like `1` (which is technically an integer)? I'm not sure if this would cause issues when writing to the TSFile — do you know if it can handle consecutive values with slightly different types like that?
> 
> If this turns out to be a potential problem, there's another approach you could consider: checking the event schema directly using the `Event` object:
> 
> ```java
> event.getSchemaInfo().getEventSchema()
> ```
> 
> This gives you a consistent view of the expected types.
> 
> Also, if you're looking for type-safe access to the data, you can use:
> 
> ```java
> event.getFieldByRuntimeName("runtimeName").getAsPrimitive().getAsBoolean()
> ```
> 
> Of course, if the current map-based approach works fine with TSFile and handles mixed types well, feel free to stick with it — just wanted to mention this as another option in case it's useful.

Thanks a lot for your suggestions! I will use these suggestions to modify my code.

luoluoyuyu (approved):

dominikriemer:
Super cool that StreamPipes gets a TsFileSink, thanks @Yuukadesu !

tenthe:
Thanks for your contribution. This is really cool.

PR #3607, opened 2025-05-12
chore(deps): drop Python 3.8 & bump pandas from ->=1.5 to >=2.0 in /streampipes-client-python

yanfeng1992:
- Changed the Python version constraint from ">3.8.1,<3.12" to ">=3.9,<3.12" in pyproject.toml.
- pandas from ->=1.5 to >=2.0 in /streampipes-client-python

https://endoflife.date/numpy

image

  

### Remarks


bossenti:
Hi @yanfeng1992,

thank you for this PR 🙌🏼 

Dropping Python 3.8 is reasonable from my point of view since it reached it's end of life already last year.
in this case we also need to modify the following parts where python 3.8 is still used/referenced:
https://github.com/apache/streampipes/blob/dev/.github/workflows/pr-validation.yml
https://github.com/apache/streampipes/blob/dev/.github/workflows/pypi-deployment.yml
https://github.com/apache/streampipes/blob/dev/.github/workflows/python-docs.yml

However, `numpy` is a transitive dependency for us only, therefore we shouldn't pin it.
If we want to avoid that `1.24` is used, we should bump the minimal supported version of `pandas` instead.

bossenti (approved):

PR #3640, opened 2025-06-02
Switch Operator Processors

matthew-reed-holden:
### Purpose

Introduce Pipeline Element Processors that allow a user to implement the "Switch" operator. I have often found myself wanting to setup a pipeline that uses the "Switch" operator but have had to use a complicated set of boolean conditions to try and implement it or I eventually end up just using a JavaScript processor to use the switch statement. The processors reflect the following block.

```java
switch(String x) {
  case "test":
   do something;
   break;
  case "test 2":
   do something else;
   break;
  default:
   do this if nothing else works;
   break;
}
```

The proposed updates introduce three processors, each which have the purpose of handling a different input type. I have implemented a Boolean Input processor, String input processor, and Numerical input processor. This is done primarily for the sake of cleaner code along with simplifying the detection of the incoming event property. These three processors could be combined into one "Switch Operator Processor" if there was the introduction of some "Conditional Requirements" component.

image

image

image

image

### Remarks

I have written unit tests for the numerical processor locally but ran into issues when I transferred the updates from my custom extensions repo to the main streampipes codebase. If someone has input on the right way to write tests for streampipes or would like to write the unit tests themselves that would be welcome.


tenthe:
Hi @heisenbergs-uncertainty,

Thanks a lot for the PR — these are great new processing elements, and your description makes it easy to review.

* For unit tests, you can use the `ProcessingElementTestExecutor` helper class.

* See `TransformToBooleanProcessorTest` or `NumericalFilterProcessorTest` for examples.

Would you like to add some unit tests for your elements?
Feel free to reach out if you have any questions or suggestions for improving the test framework.

Hope this helps!

matthew-reed-holden:
> Hi @heisenbergs-uncertainty,
> 
> Thanks a lot for the PR — these are great new processing elements, and your description makes it easy to review.
> 
> * For unit tests, you can use the `ProcessingElementTestExecutor` helper class.
> * See `TransformToBooleanProcessorTest` or `NumericalFilterProcessorTest` for examples.
> 
> Would you like to add some unit tests for your elements? Feel free to reach out if you have any questions or suggestions for improving the test framework.
> 
> Hope this helps!

Hey @tenthe! Thank you for the testing recommendation, that was super useful and made writing the tests much simpler. Props to whoever wrote that suite!

I have gone ahead and added unit tests for the processors and made quite a few updates that they helped me catch. From my testing it should be good to go at this point.

streampipes-extensions/streampipes-processors-transformation-jvm/src/main/java/org/apache/streampipes/processors/transformation/jvm/processor/switchoperator/AbstractSwitchOperatorProcessor.java
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  /**
+   * Parses a single SwitchCaseEntry from the provided StaticPropertyExtractor.
+   * This method is abstract because the parsing logic for case value and operator varies
+   * slightly between Boolean, Numerical, and String processors.
+   * @param staticPropertyExtractor Extractor for the current switch case group.
+   * @return A SwitchCaseEntry object.
+   */
+  protected SwitchCaseEntry parseSwitchCaseEntry(StaticPropertyExtractor staticPropertyExtractor) {
```
tenthe, line 119: Can this be an abstract method?

tenthe (approved):
Hi, thanks for the changes.
I just had a minor comment, but I think the code looks good from my side.

I look forward to use the processors.
Cheers,
Philipp

streampipes-extensions/streampipes-processors-transformation-jvm/src/main/java/org/apache/streampipes/processors/transformation/jvm/processor/switchoperator/AbstractSwitchOperatorProcessor.java
matthew-reed-holden: I don't think so. Switching it to abstract is causing the classes that extend it to refuse to implement the class.

But I am still refreshing my understanding of Java so I will default to your decision.

tenthe: Thanks for your response. This was just a small point I noticed while reviewing the code. I'll trigger the GitHub Action validation now. Once it passes, we can proceed with merging your PR.

PR #3641, opened 2025-06-03
fix(#3617): Validate on pipeline start that not all properties are dimensions

### LINKED ISSUE #3617: pipeline datalake all dimentions selected break stream

0.97.0

Processing Elements

### What happened?

When creating a pipeline to persist data and the end-users selects all MEASUREMENT_PROPERTY as dimensions:
There are no values to store into influx. the user is not seeing this in the UI, and the logs are very cryptic today.
```

ERROR 1 --- [    Thread-1007] o.a.s.w.s.r.StandaloneEventSinkRuntime   : RuntimeException while processing event in org.apache.streampipes.sinks.internal.jvm.datalake.DataLakeSink

java.lang.IllegalArgumentException: Expecting a positive number for fields size
```

### How to reproduce?

create data-lake and select all for dimensions

### Expected behavior

U should not be able to save the data lake settings

_No response_

None

tenthe:
### Purpose

This PR prevents users from storing only dimension properties to the data lake. If attempted, an error message (see screenshot) is shown.

**Note:**
The issue stems from InfluxDB limitations. If we switch databases, this fix can be removed.

Bildschirmfoto 2025-06-03 um 16 04 20

### Remarks


tenthe:
There seems to be a logical inconsistency with the current handling of `dimensions_selection`. In Connect, users can mark numerical values as dimension properties. However, in the Data Lake sink, these numerical dimensions are filtered out from the `dimensions` list, making it impossible for users to deselect them later.

With the current PR, all numerical values are automatically unmarked as dimension properties. This could potentially break existing pipelines relying on those settings.

Given this inconsistency in StreamPipes' behavior, we should align on a consistent approach.
@dominikriemer what’s your take on this?

dominikriemer:
I'd say we need some validation in Connect to ensure that only integer, boolean and string datatypes can be marked as dimension properties?

dominikriemer (approved):

PR #3769, opened 2025-09-09
fix: Allow anonymous access to kiosk data resource

dominikriemer:
### Remarks


streampipes-rest/src/main/java/org/apache/streampipes/rest/impl/datalake/KioskDashboardDataLakeResource.java
```diff
@@ -113,4 +116,14 @@ private boolean checkMeasureNameInWidget(DataExplorerWidgetModel widget,
   public boolean hasReadAuthority() {
```
tenthe: Is this method still required or can we delete it?

PR #3888, opened 2025-10-28
fix(3887): permission objects lost after exporting and re importing assets with sanitized resource ids

### LINKED ISSUE #3887: Permission objects lost after exporting and re-importing assets with sanitized resource IDs
### Body

When exporting an asset containing resources and importing it into another StreamPipes instance, some resources may lose their associated permission objects. During export, the IDs of `Dashboards`, `DataExplorerWidgets` and `DataStreams` are sanitized. Since these IDs are referenced in the permission objects, this causes a mismatch between the actual resource ID and the permission object ID after import.

### StreamPipes Committer

tenthe:
### Purpose
This PR addresses the loss of permission objects when importing assets with sanitized resource IDs.

The fix includes two parts:

1. A migration script that restores missing permission objects in already affected installations.
2. An update to the import logic to ensure the original resource ID is used in permission objects, preventing future mismatches.

### Remarks

### DIFF
--- a/streampipes-service-core/src/main/java/org/apache/streampipes/service/core/migrations/v0980/FixImportedPermissionsMigration.java
+++ b/streampipes-service-core/src/main/java/org/apache/streampipes/service/core/migrations/v0980/FixImportedPermissionsMigration.java
@@ -54,33 +54,33 @@ public class FixImportedPermissionsMigration implements Migration {
   }
 
   private void migrateDashboardPermissions() {
-    LOG.info("Strart migrate permissions for dashboards");
+    LOG.debug("Start migrate permissions for dashboards");
     var dataExplorerDashboardStorage = StorageDispatcher.INSTANCE
         .getNoSqlStore()
         .getDataExplorerDashboardStorage();
     var dashboards = dataExplorerDashboardStorage.findAll();
     migrateResourcePermissions(dashboards);
-    LOG.info("Finished migrate permissions for dashboards");
+    LOG.debug("Finished migrate permissions for dashboards");
   }
 
   private void migrateChartsPermissions() {
-    LOG.info("Strart migrate permissions for charts");
+    LOG.debug("Start migrate permissions for charts");
     var dataExplorerWidgetStorage = StorageDispatcher.INSTANCE
         .getNoSqlStore()
         .getDataExplorerWidgetStorage();
     var charts = dataExplorerWidgetStorage.findAll();
     migrateResourcePermissions(charts);
-    LOG.info("Finished migrate permissions for charts");
+    LOG.debug("Finished migrate permissions for charts");
   }
 
   private void migrateDataStreamPermissions() {
-    LOG.info("Strart migrate permissions for data streams");
+    LOG.debug("Start migrate permissions for data streams");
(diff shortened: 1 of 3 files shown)


streampipes-service-core/src/main/java/org/apache/streampipes/service/core/migrations/v0980/FixImportedPermissionsMigration.java
```diff
@@ -0,0 +1,116 @@
...
+
+  @Override
+  public void executeMigration() throws IOException {
+    migrateDashboardPermissions();
+    migrateChartsPermissions();
+    migrateDataStreamPermissions();
+  }
+
+  private void migrateDashboardPermissions() {
+    LOG.info("Strart migrate permissions for dashboards");
```
dominikriemer:
```suggestion
    LOG.info("Start migrate permissions for dashboards");
```

streampipes-service-core/src/main/java/org/apache/streampipes/service/core/migrations/v0980/FixImportedPermissionsMigration.java
```diff
@@ -0,0 +1,116 @@
...
+    var dataExplorerDashboardStorage = StorageDispatcher.INSTANCE
+        .getNoSqlStore()
+        .getDataExplorerDashboardStorage();
+    var dashboards = dataExplorerDashboardStorage.findAll();
+    migrateResourcePermissions(dashboards);
+    LOG.info("Finished migrate permissions for dashboards");
```
dominikriemer: maybe this can also be log level debug

streampipes-service-core/src/main/java/org/apache/streampipes/service/core/migrations/v0980/FixImportedPermissionsMigration.java
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  private void migrateChartsPermissions() {
+    LOG.info("Strart migrate permissions for charts");
```
dominikriemer:
```suggestion
    LOG.info("Start migrate permissions for charts");
```

streampipes-service-core/src/main/java/org/apache/streampipes/service/core/migrations/v0980/FixImportedPermissionsMigration.java
```diff
@@ -0,0 +1,116 @@
...
+    var dataExplorerWidgetStorage = StorageDispatcher.INSTANCE
+        .getNoSqlStore()
+        .getDataExplorerWidgetStorage();
+    var charts = dataExplorerWidgetStorage.findAll();
+    migrateResourcePermissions(charts);
+    LOG.info("Finished migrate permissions for charts");
+  }
+
+  private void migrateDataStreamPermissions() {
+    LOG.info("Strart migrate permissions for data streams");
```
dominikriemer:
```suggestion
    LOG.info("Start migrate permissions for data streams");
```

dominikriemer (approved):

PR #3948, opened 2025-11-20
Fix PLC4X S7 Advanced: Inline comments break datatype parsing

YashSaxena070:
This patch fixes an issue in the PLC4X S7 Advanced adapter where inline comments (e.g. //...) after a node definition cause schema parsing to fail.

Example failing line:
Param=%DB1281:0:REAL  // comment

Previously, the parser treated the entire string (including the comment) as the datatype, leading to:
Datatype %DB1281:0:REAL // comment is not supported

Fix

The parser now strips inline // comments before processing a node definition. Pure comment lines continue to be ignored. This update does not affect existing valid configurations.

Added

Inline comment sanitization inside parseNodeDefinition

Result

Inline comments work as expected and no longer break schema parsing.


dominikriemer:
Hi @YashSaxena070 thanks for the PR!
It seems the PR includes quite a lot changes to various files. Can you please provide the fix from a clean branch so that only the issue is adressed?
Thanks

PR #3951, opened 2025-11-21
fixes: on dev small usability fixes and fixes on filter function

JHoelli:
# TODO
- [X] Small Translation Issue in API Key Generation 
- [X] Fix in v077 migration - might make sense to delete it as it results in the default asset deleted many times and the migration should not be necessary anymore 
- [X]  Multi Filter fix in Adapter, so that no Adapters show, if filtered for a not valid combination (0 Assets)
- [X] Logout does not reset the filters - current version has disadvantage that it also resets when refreshing the page 
- [X]  Add filter in Asset view
- [X] Multifilter support for Pipelines  
- [X]  Multifilter support for Dashboards
- [X]   Multifilter support for Charts


ui/src/app/profile/components/token/token-management-settings.component.html
```diff
@@ -81,8 +81,8 @@
                     {{ 'Key created' | translate }}
                 </div>
                 <div class="subsection-small">
-                    {{ 'Your new API key has been created. Please copy the key now -
-                    you won't be able to see the key again.' | translate }}
+                    <!-- prettier-ignore -->
```
tenthe: Do we need this prettier ignore? Can't we use the line break.

PR #3952, opened 2025-11-23
fix: Update Python tutorial sources to comply with CSP

ittuann:
### Purpose

When visiting the Python docs tutorial, images fail to load due to the current Content Security Policy (CSP) configuration.

For example, when opening https://streampipes.apache.org/docs/docs/python/latest/tutorials/1-introduction-to-streampipes-python-client/ the browser console shows the following error, which cause external images fail to load in the tutorials.

```
1-introduction-to-streampipes-python-client/:2740 Refused to load the image 'https://raw.githubusercontent.com/apache/streampipes/dev/streampipes-client-python/docs/img/tutorial-preparation.gif' because it violates the following Content Security Policy directive: "default-src 'self' data: blob: 'unsafe-inline' 'unsafe-eval' https://www.apachecon.com/ https://www.communityovercode.org/ https://*.apache.org/ https://apache.org/ https://*.scarf.sh/". Note that 'img-src' was not explicitly set, so 'default-src' is used as a fallback.
```

### Remarks

no
no


bossenti:
Hi @ittuann,

Thanks a lot for fixing this! 
I have two thoughts:
1. Could we use a relative link to the images to ensure we are referencing the correct version of the documentation? With the current approach, we always refer to the content in 'dev', ignoring the version of the documents.
2. Could you remove the metadata changes from the notebooks, leaving only the relevant changes for this fix? If that's too much effort, we can leave them as they are, but it would be cleaner not to include them.

ittuann:
Hi, thanks for the feedback!

I’ve removed the metadata changes.

Regarding point 1, I suspect that using local images via relative paths is an issue with the `mkdocs-jupyter` plugin.

I found an open issue `234` in the [mkdocs-jupyter](https://github.com/danielfrg/mkdocs-jupyter) repository that seems to describe the same problem.

In addition, I noticed a info during build process, and I’m not sure whether it is related to this plugin’s issue.

```
INFO    -  DeprecationWarning: Jupyter is migrating its paths to use standard platformdirs
           given by the platformdirs library.  To remove this warning and
           see the appropriate new directories, set the environment variable
           `JUPYTER_PLATFORM_DIRS=1` and then run `jupyter --paths`.
           The use of platformdirs will be the default in `jupyter_core` v6
```

I think implementing this may take some time. Maybe we can handle this in a separate PR.

bossenti:
Thank you! 

Yes I agree with you, doesn't look like to be easy to solve. 
Let's leave it like that for now :)

bossenti (approved):

PR #3958, opened 2025-11-24
feat: Remove pipeline selection from data explorer query selector

dominikriemer:
### Purpose

To simplify user experience, remove the pipeline selection option from the data explorer so that only datasets need to be selected.

### Remarks


tenthe:
Hi @dominikriemer,

I really like this change, as it simplifies the user flow.

I have a question regarding existing charts:
How does it affect existing charts? Do we need a migration for them?

I’m currently working on the permissions. This change also allows us to remove the permissions for pipleline reads for chart users and chart admins, which helps eliminate the dependency on pipelines.

Regarding the E2E tests, I suggest running the entire test suite on this branch as well, since there may be breaking changes that wouldn’t be caught by the smoke tests alone.

tenthe (approved):

PR #3991, opened 2025-11-28
Popup to ensure saving when creating a new chart

JHoelli:
### Remarks
image

Fixed the popup appearance for creating new charts by : 

- guarding the route 
- and setting the originalDataView to the correct default values, so that the popup is only triggered if the default values are changed
- also some minor changes to resolve missing initialization or undefined errors

### DIFF
--- a/ui/src/app/data-explorer-shared/services/chart-panel-can-deactivate-guard.service.ts
+++ b/ui/src/app/data-explorer-shared/services/chart-panel-can-deactivate-guard.service.ts
@@ -30,11 +30,7 @@ export class ChartPanelCanDeactivateGuard {
     constructor(private router: Router) {}
     checkQueryParams(queryParams: { [key: string]: any }): boolean {
         const { editMode, startDate, endDate } = queryParams;
-        return (
-            editMode &&
-            startDate &&
-            endDate
-        );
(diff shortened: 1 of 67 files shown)


ui/src/app/data-explorer-shared/services/chart-panel-can-deactivate-guard.service.ts
```diff
@@ -28,12 +28,26 @@ import { SupportsUnsavedChangeDialog } from '../models/dataview-dashboard.model'
...
     constructor(private router: Router) {}
+    checkQueryParams(queryParams: { [key: string]: any }): boolean {
+        const { editMode, startDate, endDate } = queryParams;
+        return (
+            editMode !== undefined &&
+            editMode !== null &&
+            startDate !== undefined &&
+            startDate !== null &&
+            endDate !== undefined &&
+            endDate !== null
```
tenthe:
```suggestion
            editMode &&
            startDate &&
            endDate
```

tenthe: Can you do it like that?

tenthe (changes requested):

tenthe (approved):

PR #4039, opened 2025-12-09
Refactored Prometheus Endpoints to new naming convention and functioning

JHoelli:
### Purpose
Refactored Naming and functionality to #4014  (without adding the new adapterfuntionality).

### Remarks

1. Old Prometheus Endpoints set to _LEGACY and marked as deprecated.
2. Renamed and Restructured Endpoints

Disadvantage of the current deprecation approach: more (unnecessary) data is written to prometheus, legacy endpoints duplicate the current endpoints. 


streampipes-pipeline-management/src/main/java/org/apache/streampipes/manager/health/PipelineHealthCheck.java
```diff
@@ -69,6 +69,20 @@ public void checkAndRestorePipelineElements() {
...
+        p.getName(),
+        p.getHealthStatus().toString()
+    );
+
+    pipelinesStats.updatePipelineRunningState(
+        p.getElementId(),
+        p.getName(),
+        p.isRunning()
+    );
+}
```
tenthe: Can you please check the formatting of the code

PR #4047, opened 2025-12-11
feat: write data to Datalake from Java Client

JHoelli:
### Purpose
PR adds new Endpoints to java client for getting, adding and removing data from data measures. 

1. Initialize Client 
``` java
StreamPipesClient client = StreamPipesClient
                .create(<URI>,<PORT>, StreamPipesCredentials.withApiKey(<USER>, <KEY>),true);
DataLakeResourceApi dataLakeResourceApi = client.dataLakeResourceApi();
``` 
2. Get Data from Measure 
``` java
dataLakeResourceApi.get(<measureName>, new HashMap<>());
``` 
HashMap takes additional filtering arguments, e.g.  startDate or endDate 
4. Change/add Data to Measure
``` java
var data_man = List.of( Map.of(
                "timestamp", 1765290723327L,
            "density", 10.0,
            "mass_flow", 10.0,
            "sensorId", "b",
            "sensor_fault_flags", false,
            "temperature", 10.0,
            "volume_flow", 10.0
        ));
var tags = new TreeMap<String, String>();
tags.put("sensorId","a");
 SpQueryResult queryResult = null;
 try{

   queryResult = SpQuerySerializer.processEventDataWithTags(data_man, tags);}
 catch(Exception e){
    System.out.println(e.getMessage());
 }
dataLakeResourceApi.update(<measureName>, queryResult, <schemaMismatch>); 
``` 

Overwrites if timestamp exists, otherwise generate a new entry. Tags can be null but need to be set to enable influx groupBy.
6. Delete Data 
``` java
dataLakeResourceApi.delete(<measureName>,<startTimestamp>,<endTimestamp>);
``` 


streampipes-client/src/main/java/org/apache/streampipes/client/http/DeleteRequest.java
```diff
@@ -65,6 +65,9 @@ protected Request makeRequest(Serializer<K, V, T> serializer) {
 
   @Override
   protected T afterRequest(Serializer<K, V, T> serializer, HttpEntity entity) throws IOException {
+    if (entity == null || entity.getContentLength() == 0) {
```
tenthe: Why is this required? Could returning null break downstream functionality?

tenthe (approved):

PR #4067, opened 2025-12-17
Fix: ensure setuptools for mkdocs

JHoelli:
Poetry > 2.0 no longer automatically installs setup tools , however they are necessary to build the mkdocs. 

Added setuptools to poetry install -docs and as backup an install in the workflow.


bossenti (commented):
Thank you for cleaning this up @JHoelli 🙏
The "ensure setuptools is installed" step in the pipeline shouldn't be necessary as it is now included in our docs group that is installed in the step before, right?

JHoelli:
Hi @bossenti,

yes you are right, the changes to the project.toml are sufficient. I just removed it.

PR #4074, opened 2025-12-19
feat: Introducing Dataset Permissions 

JHoelli:
### Purpose

Intoducing first simple permissions for Dataset.

#### Assumptions 
- A newly created dataset inherits the permissions of the pipeline admin who created it.
- Permissions at the dataset level are user-specific, as in Connect and Pipelines.
- Option to add additional permissions (similar to e.g., Connect) 

#### Concept 

| Role                | Read | Write |
|---------------------|---|---|
| Admin               | X | X |
| Asset Admin         | - | - |
| Asset User          | - | - |
| Connect Admin       | - | - |
| Data Explorer Admin | X | - |
| Data Explorer User  | X | - |
| Dashboard Admin     | X | - |
| Dashboard User      | X | - |
| Pipeline User       | X | - |
| Pipeline Admin      | X | X |

#### Example 
A dashboard user or admin, the uses a chart which on the other hand is based on a dataset, also needs to be given explicit rights on that dataset to be able to see the chart with values.

#### Backend 
- [X]  Permission for DataLakeMeasureResources
- [X] Permissions for DataLakeResources

#### UI
- [X] Section Dataset Permission, so that everyone can see the datasets assigned 
- [X] Add Option Share Dataset in the Dataset Section 

#### Migration 
- [X] Migrate all existing measures with new permission (--> setting them to public)

#### Testing 
- [X] Write End2End Tests


ui/cypress/tests/userManagement/testUserRoleDataset.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+
+    function assertDatasetIsNotVisible(user: User) {
+        UserUtils.switchUser(user);
+        DatasetUtils.goToDatasets();
+        DatasetUtils.checkAmountOfDatasets(0);
+    }
+
+    function assertAlertBanner(exists: boolean) {
+        if (exists) {
+            cy.get('sp-alert-banner[type="error"]').should('exist');
```
tenthe: Can you change this to use the data-cy id? `cy.dataCy`

ui/cypress/support/utils/dataExplorer/DataExplorerBtns.ts
```diff
@@ -36,6 +36,14 @@ export class DataExplorerBtns {
         return cy.dataCy('save-dashboard-btn');
     }
 
+    public static discardDashboard() {
+        return cy
```
tenthe: Can you change this to a dataCy id?

ui/src/app/dataset/components/datalake-configuration/datalake-configuration.component.ts
```diff
@@ -110,6 +119,12 @@ export class DatalakeConfigurationComponent implements OnInit, AfterViewInit {
         ]);
         this.loadAvailableMeasurements();
         this.loadAvailableExportProvider();
+        const currentUser = this.currentUserService.getCurrentUser();
+        this.isAdmin = currentUser.roles.indexOf(UserRole.ROLE_ADMIN) > -1;
+        this.writeAccess =
+            currentUser.roles.indexOf(UserPrivilege.PRIVILEGE_WRITE_DATASET) >
+                -1 || this.isAdmin;
+        console.log(currentUser);
```
tenthe: Remove console.log

streampipes-pipeline-management/src/main/java/org/apache/streampipes/manager/permission/DataLakePermissionManager.java
```diff
@@ -0,0 +1,71 @@
...
+import org.apache.streampipes.model.pipeline.Pipeline;
+import org.apache.streampipes.model.staticproperty.FreeTextStaticProperty;
+import org.apache.streampipes.storage.api.IPermissionStorage;
+import org.apache.streampipes.storage.management.StorageDispatcher;
+
+import java.util.Optional;
+
+public class DataLakePermissionManager {
+
+  //TODO Could also be static
```
tenthe: Can you remove this TODO

streampipes-pipeline-management/src/main/java/org/apache/streampipes/manager/permission/DataLakePermissionManager.java
```diff
@@ -0,0 +1,71 @@
...
+private static final String DATALAKE_APP_ID =
+        "org.apache.streampipes.sinks.internal.jvm.datalake";
+private static final String DB_MEASUREMENT = "db_measurement";
+
+  public void makeAndPersistPermission(Pipeline pipeline,
+                                   String ownerSid) {
+  pipeline.getActions().stream()
+      .filter(DataSinkInvocation.class::isInstance)
+      .map(DataSinkInvocation.class::cast)
```
tenthe: Can you check if this cas is redundant?

streampipes-pipeline-management/src/main/java/org/apache/streampipes/manager/pipeline/PipelineManager.java
```diff
@@ -82,6 +83,8 @@ public static String addPipeline(
 
     Permission permission = new PermissionManager().makePermission(pipeline, principalSid);
     getPermissionStorage().persist(permission);
+    
+    new DataLakePermissionManager().makeAndPersistPermission(pipeline,principalSid);
```
tenthe:
```suggestion
    new DataLakePermissionManager().makeAndPersistPermission(pipeline, principalSid);
```

streampipes-rest/src/main/java/org/apache/streampipes/rest/impl/datalake/AbstractDataLakeResource.java
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  /**
+   * required by Spring expression
+   */
+  public boolean hasWriteAuthority() {
+    return isAdminOrHasAnyAuthority(DefaultPrivilege.Constants.PRIVILEGE_WRITE_DATASET_VALUE);
+  }
+
+  public boolean checkPermission(String measurementName,
```
tenthe: Does it make sense to rename it to something like `checkPermissionByName`?

streampipes-rest/src/main/java/org/apache/streampipes/rest/impl/datalake/AbstractDataLakeResource.java
```diff
@@ -0,0 +1,76 @@
...
+    var spPermissionEvaluator = new SpPermissionEvaluator();
+    var authentication = SecurityContextHolder.getContext()
+        .getAuthentication();
+    return spPermissionEvaluator.hasPermission(
+        authentication,
+        measurementName,
+        permission);
+  }
+
+  public boolean checkDatasetPermission(String measurementId,
```
tenthe: Same here `checkPermissionById`?

streampipes-service-core/src/main/java/org/apache/streampipes/service/core/migrations/v099/CreateDatasetPermissionMigration.java
```diff
@@ -0,0 +1,68 @@
...
+
+  @Override
+  public void executeMigration() throws IOException {
+    dataLakeStorage.findAll().forEach(measure -> {
+      var existingPermission = permissionStorage.getObjectPermissions(List.of(measure.getMeasureName()));
+      if (existingPermission.isEmpty()) {
+        permissionResourceManager.createDefault(
+            measure.getMeasureName(),
+            DataLakeMeasure.class,
+            null,
```
tenthe: For the measurement owner: would it be possible to check whether there is a pipeline writing to this measurement? If so, we could use the pipeline owner as the default; otherwise, we would keep the value as null. What do you think?

tenthe (changes requested):
Hi @JHoelli,

thanks a lot for the PR, this is a really cool new functionality.
I tested it locally and everything worked as expected, including the migration.

I only have a few minor code comments.

tenthe:
Hi @JHoelli,

I just realized we need to change the logic for setting data lake permissions. Currently, this is handled in `PipelineManager.addPipeline`, but this is not the only place where a measurement can be created. A measurement can also be created via the API, so this logic should be moved to the REST interface, `DataLakeMeasureResource`.

streampipes-pipeline-management/src/main/java/org/apache/streampipes/manager/permission/DataLakePermissionManager.java
JHoelli: necessary as not all Actions are DataSinks and we only want to obtain the Sinks

JHoelli:
# Current Issue

When moving permission management for datasets from the Pipeline Manager to the DataLakeResource, we are facing an issue where the principal ID is lost if a data lake is created from a pipeline. This occurs because the getAuthenticatedUserSid() method only returns the user ID of the service principal (sp-service-client), which results in incorrect ownership assignments.

On the other hand, when a data lake is created via API, the correct ownership is assigned, as the user token is properly used.

Creating a data lake via API results in the correct ownership due to the provided user token.

# Possible Solutions 

## Move the Measurement Creation to the DataSinkInvocation 

**Idea:** Split the measurement creation and update process. Only create the measurement and pipeline permissions within the Pipeline Manager.

**Pros:**

- Separates concerns, keeping pipeline management cleaner.

- Allows permissions to be set specifically for the pipeline context (using the SchemaManager not the ClientAPI).

**Cons:**

- more complex flow management.

- Reusability

## Add an additional Header 

**Idea:** Introduce a new HTTP header to specify if the authenticated user is an admin. If the user is an admin, allow the system to use the user ID from the token (via TokenAuthenticationFilter) instead of the service principal ID or introduce an  'onbehalfof'.

Reasoning: If a user already has admin rights, setting another user reduces those rights.

**Pros:**

- Reusable solution for other use cases.

- Ensures the correct user ownership.

**Cons:**

- may be harder to understand.
- no other usage  of such logic.


## Create Permissions in the Pipeline Manager and only check if Permissions are available in the DataMeasureResource 

**Idea:** Revert to the old approach: create the permissions in the Pipeline Manager with the authenticated user. In the DataMeasureResource, simply check whether permissions already exist. If not, use the currently authenticated user to set the permission.

**Pros:**

- Easy to implement.

- Ensures functionality with the StreamPipes client and core systems.

**Cons:**

- Permissions are managed in two separate places, making the implementation less clean.

- Reduces reusability, as permission management is not centralized.

tenthe:
Merged as part of #4099

Issue #4113, opened 2026-01-19
Extend MS-Teams  sink error handling


obermeier:
### Body

 I found some possible improvements of the MS Teams sink.

-  Close HTTP responses to prevent hanging connections / connection leaks
- Add retry logic (with backoff) when the MS Teams webhook rate limit is hit (HTTP 429)
- Throw an exception when message delivery fails (non-2xx response after retries)

### StreamPipes Committer

I acknowledge that I am a maintainer/committer of the Apache StreamPipes project.

PR #4114, opened 2026-01-19
#4113  Msteams sink error handling

obermeier:
I found some possible improvements of the MS Teams sink.

 -   Close HTTP responses to prevent hanging connections / connection leaks
 -   Add retry logic (with backoff) when the MS Teams webhook rate limit is hit (HTTP 429)
 -   Throw an exception when message delivery fails (non-2xx response after retries)


obermeier:
Based on old version

PR #4130, opened 2026-01-26
fix(#4120): Add bearer token authentication options to HttpStreamProtocol

### LINKED ISSUE #4120: Add custom header configuration to HTTP stream adapter
### Discussed in https://github.com/apache/streampipes/discussions/4119

<div type='discussions-op-text'>

<sup>Originally posted by **feribudi2** January 20, 2026</sup>
How do I consume a REST API from an external server that requires a Bearer Token? What connector do I need to use?</div>

tenthe:
### Purpose

Adds optional HTTP bearer authentication to the HttpStreamProtocol adapter, updates locale strings, and introduces a  migrator to add the new auth alternatives while bumping the adapter version to 1.

### Remarks


dominikriemer:
Hi @tenthe 
maybe we could use a `CollectionStaticProperty` with key/value pair inputs as plain text. In this case, we would also support any other custom headers users want to add.

e.g., 

```
CollectionStaticProperty
  StaticPropertyGroup
    Header Key: FreeTextStaticProperty
    Header Value: FreeTextStaticProperty
```

Examples:
Content-type: application/json
X-API-User: username
...

dominikriemer (approved):

feribudi2:
image
How do I enter a Bearer Token? I don't see a field or menu to enter the bearer token header?

tenthe:
Hi @feribudi2,

could you confirm which StreamPipes version you are currently using? From your description, it looks like you’re on the latest release (0.98.0), is that correct?

The fix for this issue has not been included in a stable release yet. It is currently only available in the snapshot version (0.99.0-SNAPSHOT).

In this upcoming version, you will be able to provide custom request headers as part of the configuration.

grafik

feribudi2:
> Hi @feribudi2,
> 
> could you confirm which StreamPipes version you are currently using? From your description, it looks like you’re on the latest release (0.98.0), is that correct?
> 
> The fix for this issue has not been included in a stable release yet. It is currently only available in the snapshot version (0.99.0-SNAPSHOT).
> 
> In this upcoming version, you will be able to provide custom request headers as part of the configuration.
> 
> grafik

I'm using version 0.98, okay, I'll try version 0.99.

Thanks for the info.

Issue #3621, opened 2025-05-22
plc4x s7 - register input- advanced: fault in comment parsing


jormus:
### Apache StreamPipes version

0.97.0

### Affected StreamPipes components

Connect

### What happened?

When defining nodes in plc4x s7 advanced, u cannot put comments on the same line as the node definition:

example config triggering fault when trying to define schema
`
H_AnInput_HighLimit=%DB1281:0:REAL	// High Limit For Scaling Analogue Input
`

results in:
`
Datatype %DB1281:0:REAL // High Limit For Scaling Analogue Input is not supported
`

### How to reproduce?

works:
```
// High Limit For Scaling Analogue Input
H_AnInput_HighLimit=%DB1281:0:REAL
```
works not:
```
H_AnInput_HighLimit=%DB1281:0:REAL	// High Limit For Scaling Analogue Input
```

### Expected behavior

// should be ignored

### Are you willing to submit a PR?

None

tenthe:
Hi @jormus,

Thank you very much for reporting this issue.

I've reclassified it as an enhancement, as the configuration syntax is already described in the header. That said, I really appreciate the suggestion of supporting comments at the end of each line—it's a great idea for improving clarity.

I've also marked this as a good first issue, since the change should be relatively straightforward to implement. If anyone is interested in contributing, this would be a great place to start.

Please don't hesitate to share any further suggestions that could help improve the usability of the software.

jormus:
What would be really cool if line comments are picked up by schema definitions and mapped to the description field. 
For self service this would be great, I plan on automating and versioning the adapters / pipelines
In the meantime, , I keep breaking the platform ;-)

@tenthe I have quite a few improvements and ideas tbh. But for now, German engineering at its best ;-)

tenthe:
Hi @jormus,
using comments in the description field is a great idea!

For automation and versioning, have you seen these features?

* [Adapters as Code](https://streampipes.apache.org/docs/use-programmatically-create-adapters/)
* [Pipelines as Code](https://streampipes.apache.org/docs/use-programmatically-create-pipelines/)

They should help support your planned implementation.

jormus:
@tenthe 
yes, have seen it exists, not yet used, is it available in the python sdk?
Then the effort to pick up repo merges and pushing them to streampipes sounds interesting.
Still testing the stack.
I guess I should consult the company, lots of questions tbh.
My first question, is why adapters and devices is not split out. because one source can create n amount of adapters
[...]

PR #4142, opened 2026-01-31
feat(plc4x-s7): support inline // comments in PLC4X S7 advanced configuration

alisha-1000:
This PR adds support for inline `//` comments in the PLC4X S7 advanced code block configuration.

Previously, comments placed at the end of a configuration line were treated as part of the PLC datatype, which caused parsing errors such as:
`Datatype %DB1281:0:REAL // comment is not supported`.

With this change, inline comments are stripped before parsing key/value pairs, while preserving support for full-line comments and existing configurations.

### DIFF
--- a/streampipes-extensions/streampipes-connectors-plc/src/main/java/org/apache/streampipes/extensions/connectors/plc/adapter/s7/config/ConfigurationParser.java
+++ b/streampipes-extensions/streampipes-connectors-plc/src/main/java/org/apache/streampipes/extensions/connectors/plc/adapter/s7/config/ConfigurationParser.java
@@ -49,20 +49,27 @@ public class ConfigurationParser {
       // Remove leading and trailing whitespace
       line = line.trim();
 
-      // Skip comments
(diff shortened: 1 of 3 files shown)


tenthe:
Hi @alisha-1000,

thanks a lot for this PR!
This looks good to me.

If you want, you could also add some unit tests for the new functionality in `ConfigurationParserTest`.

Fixes #3621

alisha-1000:
Hi @tenthe,
Thanks for the suggestion! I’ve added a unit test to cover inline `//` comments in ConfigurationParserTest.

tenthe:
Hi @alisha-1000,

somehow the changes from this PR #4147 ended up here.
Can you please remove it so there will be no merge conflict?

alisha-1000:
Hi @tenthe,

Thanks for pointing that out.
I’ve reverted the unintended UI-related commit, so this PR now only affects the PLC4X S7 inline comment parsing.

Appreciate the review!

Issue #4112, opened 2026-01-19
Update ShepherdService to support new adapter generation workflow


tenthe:
### Body

The ShepherdService used for tutorials is not compatible with the new adapter generation workflow and must be updated accordingly.

### StreamPipes Committer

I acknowledge that I am a maintainer/committer of the Apache StreamPipes project.

alisha-1000:
Hi @tenthe, thanks for opening this issue.

I’d like to take a look at updating the ShepherdService to align with the new adapter generation workflow.

Before starting, I wanted to clarify a few things to ensure I’m aligned with the intended approach:

Are there specific changes in the adapter generation workflow that ShepherdService currently does not support?

Is the incompatibility mainly related to UI flow, API changes, or both?

Are there any existing references or components that already support the new workflow and could be reused here?

Once clarified, I’d be happy to work on a PR.

tenthe:
Hi @alisha-1000,

thanks for the clarification questions.

In the adapter creation workflow, a new step was added that allows users to transform the event using a script. The existing adapter creation tour has not been updated to cover this new functionality.

Please run the tour until it fails (see screenshot):

<img width="632" height="354" alt="Image" src="https://github.com/user-attachments/assets/b1d94a50-b2c2-499e-a946-b75dc0fb66f2" />

The relevant code can be found in `ui/src/app/services/tour/adapter-tour.constants.ts`. When updating the tour, please use the existing `data-cy` selectors where possible. These are used by Cypress for automated tests and can also be used to identify UI elements.

If you have any questions, feel free to ask.

alisha-1000:
Hi @tenthe,

I’ve opened a PR that updates the adapter creation tour to include the new event transform script step and align it with the current adapter generation workflow:
PR: #4147
The change updates `ui/src/app/services/tour/adapter-tour.constants.ts` and reuses existing `data-cy` selectors as suggested.

Happy to make any adjustments if needed.

PR #4147, opened 2026-02-02
[#4112] Fix adapter creation tour for event transform script step

alisha-1000:
### Purpose

This PR updates the adapter creation Shepherd tour to align with the new adapter generation workflow.

A new step was introduced in the adapter creation process that allows users to transform incoming events using a script. The existing adapter tour did not include this step, which caused the tour to break during execution.

This change adds the missing tour step and updates the flow so the tutorial proceeds correctly with the current UI.

### Remarks

Only the Shepherd tour configuration was updated.

No functional UI or backend logic was modified.

Existing data-cy selectors were reused to remain compatible with Cypress tests.

### DIFF
--- a/ui/src/app/services/tour/adapter-tour.constants.ts
+++ b/ui/src/app/services/tour/adapter-tour.constants.ts
@@ -99,6 +99,14 @@ export default {
             },
             {
                 stepId: 'step-11',
+                title: 'Transform events using a script',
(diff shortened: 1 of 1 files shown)


dominikriemer (approved):

dominikriemer:
Thanks @alisha-1000 !

alisha-1000:
@dominikriemer It looks like most remaining checks are cancelled due to workflow approval / runner availability.
I don’t see any test failures related to the PR changes.
Could you please advise or approve the workflows when convenient?
Thanks!

PR #4155, opened 2026-02-09
Fix user edit dialog requiring username for existing users

### LINKED ISSUE #2681: User edit dialog expects username

UI

### What happened?

When editing a user in the user management, the dialog expects the user name to be provided.

image

This is an issue since the username is never present when the user was created through self-registration.

### How to reproduce?

_No response_

### Expected behavior

_No response_

_No response_

No, I don't want to submit a PR.

Jayyadav1903:
### Problem
When editing an existing user, the User Edit Dialog enforced a required validator on the username field.
This breaks editing for:
- self-registered users
- the initial admin user
because these accounts may not have a username.

### Solution
Relax validation **only in edit mode** for local user accounts:
- username is optional when editing
- if provided, it must still be a valid email
- creation behavior remains unchanged

### Implementation
- Updated `getUsernameValidators()` in `edit-user-dialog.component.ts`
- Removed `Validators.required` when:
  - `editMode === true`
  - `isUserAccount === true`
  - `provider === 'local'`

### Impact
- UI-only change
- No backend or data model changes
- No breaking behavior

### Verification
- Edit local user without username → ✅ allowed
- Create local user → username still required
- Service accounts → unchanged
- External (LDAP) users → unchanged

Fixes #2681

### DIFF
--- a/docs/architecture/DOMAIN_PROPERTY_INVENTORY.md
+++ b/docs/architecture/DOMAIN_PROPERTY_INVENTORY.md
@@ -1,0 +1,43 @@
+# Domain Property Inventory
+
+This document lists hard-coded domain properties found in the Apache StreamPipes repository, as part of the analysis for Issue #1373.
+
+## Findings
+
+| File Path | Component | Domain Property String | Context |
+| :--- | :--- | :--- | :--- |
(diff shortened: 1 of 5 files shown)


Jayyadav1903:
fix(ui): allow editing users without username in edit dialog (#2681)

PR #4167, opened 2026-02-10
fix(connect-ui): load OPC UA browse tree in edit mode without waiting for dependencies

Jayyadav1903:
### Problem

When editing an existing OPC UA adapter, the **Browse** tab remains empty.
This happens because `StaticRuntimeResolvableTreeInputComponent` waits for
dependent properties to emit completion events, which do not occur in edit mode
since adapter values are pre-filled and no change events are triggered.

As a result, the static browse tree is never initialized.

---

### Solution

This PR treats **edit mode** as a resolved configuration state:

- If `adapterId` is present (indicating edit mode), the static property
  immediately attempts to load its options from the REST API
- This bypasses waiting for dependency completion events that never fire in this scenario
- The existing dependency-based behavior for adapter creation remains unchanged

---

### Implementation Details

- Updated `StaticRuntimeResolvableTreeInputComponent.ngOnInit`
- Added a guard to trigger `loadOptionsFromRestApi()` when `adapterId` is present,
  even if `dependsOn` is defined
- No backend logic or adapter configuration was changed

```ts
if (
    this.staticProperty.nodes.length === 0 &&
    (
        !this.staticProperty.dependsOn ||
        this.staticProperty.dependsOn.length === 0 ||
        this.adapterId
    )
) {
    this.loadOptionsFromRestApi();
}


Jayyadav1903:
Hi @tenthe,

I’ve opened a PR that addresses this issue. The fix ensures the OPC UA Browse tree is loaded immediately in **edit mode** by treating an existing adapter configuration (`adapterId` present) as resolved, instead of waiting for dependency completion events that never fire in this scenario.

Adapter creation behavior remains unchanged, and the existing Cypress test should pass once run against a live backend.

Let me know if you’d like this handled differently.

Thanks!

PR #4223, opened 2026-03-05
chore: Simplify code to verify pipeline elements

dominikriemer:
### Purpose

Cleanup the element verification process when storing new pipeline elements and adapters.

### Remarks


streampipes-pipeline-management/src/test/java/org/apache/streampipes/manager/verification/TestTypeExtractor.java
```diff
@@ -0,0 +1,114 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+
+public class TestTypeExtractor {
```
tenthe: For the other unit tests, we used Test as a suffix. Should we change it here for consistency?
I don’t think we have documented this anywhere yet, where would be a good place to document it?
Should we enforce this with Checkstyle, or is it sufficient to keep it as a best practice?

tenthe (changes requested):

tenthe (approved):

PR #4240, opened 2026-03-11
feat: Full integration of filter feature for table widget, excel like filte…

Nguyen-Bang:
I added one features to the table widget. The tables can now be filtered just like in excel. That means dynamic filtering. One can input a search term to filter by specific values (by just checking the values that one needs). Furthermore, the filters also provide an advance option just like in excel. That means numbers have number base filter options and everything else has text base filter options. 

image
image
image
image


tenthe (approved):
Hi @Nguyen-Bang,
Thanks for this PR! I tested it locally and everything worked as expected.
Really appreciate the addition of the E2E test as well.

PR #4284, opened 2026-03-23
Feat:Improved dark theme implementation

Nguyen-Bang:
Functionality changes: 
 -removed "save color schema" button in profile for more intuitive handling
 -all chart types (table, pie, heatmap etc.) have default color on initialization (before it was on custom) to produce better visualization

Appearance change to mostly match Angular Material 3 standard (removed the use of manual light-dark() implementation):
 -changed background to dark grey (before black) for better contrast 
 -removed any hardcoded colors that don't respect theme switching
 -exception: hard coded color of the menu-group-title (words "Analytics", "Visualization", "Management"), since they are on a green background and difficult to read with the default color schema

The first and last 3 pictures are the before, all others are the after
Screenshot from 2026-03-23 18-42-45
Screenshot from 2026-03-23 18-42-16
Screenshot from 2026-03-23 18-42-02
Screenshot from 2026-03-23 18-41-16
Screenshot from 2026-03-23 18-41-06
Screenshot from 2026-03-23 18-27-52
Screenshot from 2026-03-23 16-58-51
Screenshot from 2026-03-23 16-58-41
Screenshot from 2026-03-23 16-58-04

### DIFF
--- a/ui/src/app/core/components/iconbar/iconbar.component.scss
+++ b/ui/src/app/core/components/iconbar/iconbar.component.scss
@@ -284,5 +284,5 @@
 
 :host-context(.dark-mode) .menu-group-title,
 :host-context(.dark-mode) .chevron {
-    color: #120c47 !important;
+    color: var(--color-menubar-text-menu-group) !important;
 }
--- a/ui/src/app/core/components/iconbar/iconbar.component.scss
+++ b/ui/src/app/core/components/iconbar/iconbar.component.scss
@@ -281,3 +281,8 @@
         --mb-width: 248px;
     }
 }
+
+:host-context(.dark-mode) .menu-group-title,
+:host-context(.dark-mode) .chevron {
+    color: #120c47 !important;
+}
--- a/ui/src/app/profile/components/general/general-profile-settings.component.html
+++ b/ui/src/app/profile/components/general/general-profile-settings.component.html
@@ -116,16 +116,6 @@
                         >
                     </mat-radio-group>
                 </sp-form-field>
-                <div>
-                    <button
-                        mat-button
-                        mat-flat-button
-                        color="accent"
-                        (click)="updateAppearanceMode()"
-                    >
-                        {{ 'Save color schema' | translate }}
-                    </button>
-                </div>
             </sp-split-section>
         </div>
     }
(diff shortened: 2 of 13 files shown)


ui/src/app/core/components/iconbar/iconbar.component.scss
```diff
@@ -281,3 +281,8 @@
         --mb-width: 248px;
     }
 }
+
+:host-context(.dark-mode) .menu-group-title,
+:host-context(.dark-mode) .chevron {
+    color: #120c47 !important;
```
dominikriemer: To support different color schemes, we can add a variable in deployment/theme/_custom-variables

ui/src/app/profile/components/general/general-profile-settings.component.html
```diff
@@ -116,16 +116,6 @@
                         >
                     </mat-radio-group>
                 </sp-form-field>
-                <div>
```
dominikriemer, line 119: is there a reason why we don't keep the save button?

ui/src/scss/sp/sp-theme.scss
```diff
@@ -43,39 +42,54 @@ html {
             density: -2,
         ),
         $overrides: (
-            primary: light-dark(var(--color-primary), var(--color-primary-dark)),
```
dominikriemer, line 46: I'd suggest to keep the light-dark pattern here since some user-defined colors that are different to the StreamPipes colors need to be adjusted between light and dark mode.

dominikriemer (changes requested):

ui/src/app/profile/components/general/general-profile-settings.component.html
Nguyen-Bang: Not particularly. 
I just thought it would be more intuitive. It happened several times that I switched the theme and didn't save it (just a small inconvenience). The "problem" is that clicking on the opposite theme immediately changes the visual without clicking the save button, so to the user it already feels like the theme is changed even though it is not saved yet. 
As I see it, there are no real downsides to having or not having the button. 
I feel like if we keep the save button, maybe we shouldn't immediately change the visuals but only change them the moment the user clicks on save

ui/src/app/core/components/iconbar/iconbar.component.scss
Nguyen-Bang: Hi @dominikriemer ,
I have changed this now.

dominikriemer (approved):

PR #4305, opened 2026-03-26
fix: resolve Python Client dependency issues for Python 3.13 support

ittuann:
### Purpose

In the current `mkdocs-material` (v9.3.2), the `regex` dependency is strictly pinned to `>=2022.4,<2023.0`. These older versions of regex do not provide pre-compiled wheels for Python 3.13, leading to installation errors or requiring a complex build environment.

This PR upgrades mkdocs-material to v9.4.2, which fix the strict upper bound on the regex dependency (see [Issue#6068](https://github.com/squidfunk/mkdocs-material/issues/6068)), resolving the dependency conflicts and build failures encountered in Python 3.13 environments.

### Remarks

no
no


tenthe (approved):

ittuann:
rebase to resolve conflicts in poetry.lock caused by dependabot

PR #4317, opened 2026-03-27
fix: resolve CI Pull Request Labeler error for fork PRs

ittuann:
### Purpose

The `actions/labeler` requires `pull-requests: write` permissions to apply labels. However, When a pull request is submitted from a forked repository, workflows triggered by `pull_request` are get minimal, read-only permissions.
This let the Labeler action cannot apply labels to PRs coming from forks.

For more details, see: https://github.com/actions/labeler/issues/121#issuecomment-1036490254
Additionally, the example workflow in the Labeler README also uses `pull_request_target`: https://github.com/actions/labeler#create-workflow

While I noticed this trigger was previously changed in https://github.com/apache/streampipes/pull/4149
Regarding security, I think using `pull_request_target` in this workflow is safe because this CI workflow does not check out any external code (no `actions/checkout`). The workflow simply runs in the context of the base repository rather than the fork, safely granting the necessary write permissions to apply labels to the PR without executing any untrusted code.

### Remarks

no
no

### DIFF
--- a/.github/workflows/pr-labeler.yml
+++ b/.github/workflows/pr-labeler.yml
@@ -17,7 +17,7 @@ name: "Pull Request Labeler"
 run-name: Pull request labeler for ${{ github.ref_name }}
 
 on:
-  pull_request:
+  pull_request_target:
 
 jobs:
   label-pr:
--- a/.github/workflows/pr-labeler.yml
+++ b/.github/workflows/pr-labeler.yml
@@ -34,6 +34,5 @@ jobs:
     - uses: actions/labeler@v6
       with:
         configuration-path: .github/pr-labeler-config.yml
-        repo-token: "${{ secrets.GITHUB_TOKEN }}"
         # Remove labels when matching files are reverted
         sync-labels: true


dominikriemer:
Hi @ittuann unfortunately the ASF policy [1] does not allow to use `pull_request_target` on any workflow that exports a secret and the action uses `repo-token: "${{ secrets.GITHUB_TOKEN }}"`.

What do you think, should we just remove PR labeling then for PRs?

[1] https://infra.apache.org/github-actions-policy.html

ittuann:
Thanks for the clarification @dominikriemer 

Since `actions/labeler@v6` provides the default value for `repo-token`, we can remove the explicit `GITHUB_TOKEN` using, while retaining the `pull_request_target` event. https://github.com/actions/labeler?tab=readme-ov-file#inputs

I've also checked how other Apache repositories handle this, and they follow a similar pattern. https://github.com/apache/avro/pull/3659

I personally prefer keeping the labeling feature as it is intuitive and helpful for reviewing and tracking PRs.

If this is still restricted, we can alternatively restrict the execution permissions like so:

```yaml
- if: |  
  github.event.pull_request.author_association == 'MEMBER' || 
  github.event.pull_request.author_association == 'OWNER'
```

dominikriemer (approved):

PR #4341, opened 2026-04-01
fix: pin CSV import actions to bottom

tenthe:
### Purpose

Adjusted `CsvImportDialogComponent` so the dialog content fills the slide-in panel height and the navigation/footer stays pinned to the bottom, matching the expected wizard layout. 

Remove unused enums in `page-name.enum.ts`

### Remarks


ui/src/app/dataset/dialog/csv-import-dialog/csv-import-dialog.component.scss
```diff
@@ -17,19 +17,23 @@
...
     gap: 18px;
-    max-height: 78vh;
+    min-height: 0;
+    max-height: none;
     overflow: auto;
 }
 
 :host {
     display: block;
+    height: 100%;
```
dominikriemer: We don't have to specify custom css for the dialog. We can use `sp-dialog-actions` or `sp-dialog-content` similar to other dialogs

dominikriemer (approved):

PR #4342, opened 2026-04-01
feat(#4314): add streampipesclient to javascript transformation context

### LINKED ISSUE #4314: Add StreamPipesClient to JavaScript transformation context
### Body

The JavaScript connect transformations currently receives an empty context. This context should be enriched by adding an instance of a `StreamPipesClient`.
The goal is to make the connect transformation much more powerful by allowing access to the client and be able to read and write data via the API. To ensure a safe usage of the client it has to be scoped with `.onBehalfOf(userId)` to limit the access to the users permissions.

### StreamPipes Committer

SvenO3:
### Purpose

Fixes #4314

### Remarks


streampipes-connect-management/src/main/java/org/apache/streampipes/connect/management/management/GuessManagement.java
```diff
@@ -100,6 +104,11 @@ public SampleData getSampleData(AdapterDescription adapterDescription)
   }
 
   public AdapterDescription transformSampleData(AdapterDescription adapterDescription) throws AdapterException {
```
tenthe: I think we can migrate this to one function. The current approach might not work for compact adapters.

tenthe (changes requested):

streampipes-connect-management/src/main/java/org/apache/streampipes/connect/management/management/GuessManagement.java
SvenO3: It's fixed but I needed to change the `AdapterModelGenerator` interface for it

tenthe (approved):

PR #4360, opened 2026-04-08
feat: Sandbox execution of untrusted groovy code

SvenO3:
### Purpose

This PR makes Groovy-based transformation scripts more secure by introducing a dedicated sandbox that limits the access to certain operation. The sandbox policy for example blocks direct access to the file system, process APIs, JVM exit paths and networking classes. The `GroovyScriptEngine` now compiles scripts through that sandbox while preserving the existing `input`, `out`, and `ctx` execution model. It also includes a detailed error handling to show the user which operations are restricted.
Moreover a bug with the script context was fixed that Groovy can now also use the `StreamPipesClient` via `ctx.client()`.

### Remarks

### DIFF
--- a/streampipes-connect-transformer-groovy/pom.xml
+++ b/streampipes-connect-transformer-groovy/pom.xml
@@ -44,6 +44,11 @@
             <groupId>org.apache.groovy</groupId>
             <artifactId>groovy</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/streampipes-connect-transformer-groovy/pom.xml
+++ b/streampipes-connect-transformer-groovy/pom.xml
@@ -41,6 +41,11 @@
             <version>0.99.0-SNAPSHOT</version>
         </dependency>
         <dependency>
+            <groupId>org.apache.streampipes</groupId>
+            <artifactId>streampipes-client-api</artifactId>
+            <version>0.99.0-SNAPSHOT</version>
+        </dependency>
+        <dependency>
             <groupId>org.apache.groovy</groupId>
             <artifactId>groovy</artifactId>
         </dependency>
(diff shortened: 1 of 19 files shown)


streampipes-extensions-management/src/main/java/org/apache/streampipes/extensions/management/connect/adapter/ScriptContextResolver.java
```diff
@@ -28,16 +30,32 @@ public class ScriptContextResolver {
 
   public Context resolve(AdapterDescription adapterDescription) {
     var userId = getUserId(adapterDescription);
+    var language = adapterDescription.getTransformationConfig().getLanguage();
+    return resolve(userId, language);
+  }
+
+  public Context resolve(String userId, String language) {
```
tenthe, line 37: Can this be changed to private?

streampipes-connect-transformer-groovy/pom.xml
```diff
@@ -44,6 +44,23 @@
...
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.streampipes</groupId>
+            <artifactId>streampipes-client-api</artifactId>
+            <version>0.99.0-SNAPSHOT</version>
+            <scope>compile</scope>
```
tenthe: I think the scope can be removed here

streampipes-extensions-management/pom.xml
```diff
@@ -123,6 +123,12 @@
             <artifactId>mockito-core</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.streampipes</groupId>
+            <artifactId>streampipes-connect-transformer-groovy</artifactId>
+            <version>0.99.0-SNAPSHOT</version>
+            <scope>compile</scope>
```
tenthe: Do we need the scope here?

tenthe (changes requested):
Hi @SvenO3,
Thanks for the changes. I added some minor comments. 
What do you think about them?
Cheers,
Philipp

streampipes-extensions-management/src/main/java/org/apache/streampipes/extensions/management/connect/adapter/ScriptContextResolver.java
SvenO3: This is also needed by `GuessManagement` when the adapter isn't created yet and therefore the userId can't be retrieved by the `PermissionResourceManager`

tenthe (approved):

PR #4398, opened 2026-04-20
fix(#4374): Add encodeURIComponent to data lake REST requests

### LINKED ISSUE #4374: Fix Escaping Issues for Dataset and Adapter Names Containing Square Brackets
### Body

Datasets or adapters whose names contain square brackets (`[]`) can currently cause failures in system behavior. For example, such datasets cannot be deleted because the REST request fails, and in at least one case an adapter with a name like `Test []` and a persist pipeline could not be stopped.

This indicates that there is likely an escaping or request handling problem for square brackets in names, especially in API calls or pipeline-related operations.

The root cause should be identified and fixed so that users can safely use `[` and `]` in dataset and adapter names without breaking deletion, stop actions, or other related functionality.

If full support for these characters cannot be implemented reliably, the system should at minimum validate and warn users when they try to use them in names, so that the resources can not be created.

### StreamPipes Committer

stefaniesinner:
### Purpose

This PR fixes failing HTTP requests for datasets and adapters containing square brackets. The root cause is a frontend issue where raw brackets (`[]`) are sent in URLs, which are invalid characters and get rejected by the server. The backend requires no changes since Spring Boot and the databases already handle these characters safely.

It adds `encodeURIComponent()` to REST paths in `datalake-rest.service.ts` so that the names are properly URL-encoded (e.g., `[` becomes `%5B`).

### Result
StreamPipes no longer fails when stopping or deleting datasets and adapters that have square brackets in their names.

It also resolves the data visualization error in the chart mentioned in the issue comments.

image

### Remarks

Fixes #4374

### DIFF
--- a/ui/projects/streampipes/platform-services/src/lib/apis/datalake-rest.service.ts
+++ b/ui/projects/streampipes/platform-services/src/lib/apis/datalake-rest.service.ts
@@ -93,7 +93,9 @@ export class DatalakeRestService {
 
     getMeasurementByName(name: string): Observable<DataLakeMeasure> {
         return this.http
-            .get(`${this.dataLakeMeasureUrl}/byName/${name}`)
+            .get(
+                `${this.dataLakeMeasureUrl}/byName/${encodeURIComponent(name)}`,
+            )
             .pipe(map(res => res as DataLakeMeasure));
     }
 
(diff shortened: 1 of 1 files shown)


tenthe (approved):

tenthe:
Hi @stefaniesinner,
thanks a lot for providing this PR

PR #4483, opened 2026-05-18
feat(#4430): Chart settings not correctly migrated after event schema changes

### LINKED ISSUE #4430: Chart settings not correctly migrated after event schema changes
### Body

When a user changes the event schema (e.g., via adapter configuration or pipeline), these changes are not reflected in existing charts. A mechanism is needed to handle schema updates: non-breaking changes should be applied automatically, while breaking changes should trigger a user warning. A concept for this behavior needs to be designed before implementation. Open questions include how to present this to users and where to surface it (e.g., adapter migration dialog, pipeline dialog, or a separate overview of affected charts).

### StreamPipes Committer

SvenO3:
### Purpose

This handles changes of adapters and pipelines that effect existing charts by using the following approach:
- When updating an adapter or pipeline it checks if the corresponding datasets get displayed in any charts. Then it searches if any field that is used for a visualization gets deleted and displays a warning for it. For adapter (see image 1)  the exiting migration dialog was enhanced while the pipelines (see image 2) got a new migration dialog
- If the user confirms the migration, the migration is executed and the affected chart get flagged in the chart overview and dashboards that use these charts also show a warning (see image 3)
- Now the charts need to be adjusted manually. When the chart editor opens the a warning is shown with displays the deleted fields

#### 1) Adapter Changes
image

#### 2) Pipeline Change
image

#### 3) Dashboard View
image

### Remarks


ui/src/app/chart/components/chart-view/chart-view.component.ts
```diff
@@ -28,9 +28,11 @@ import {
     ChartService,
     DataExplorerWidgetModel,
     DataLakeMeasure,
+    DatalakeRestService,
```
tenthe: Remove unused import

ui/src/app/chart/components/chart-view/chart-view.component.ts
```diff
@@ -28,9 +28,11 @@ import {
...
     EventPropertyUnion,
     FieldConfig,
     LinkageData,
+    SourceConfig,
```
tenthe: Remove unused import

tenthe (changes requested):

tenthe (changes requested):
There is just an unused import, please remove this. The PR looks good to me. Thank you

tenthe (approved):

PR #4487, opened 2026-05-19
feat: Workflow consistency for Dashboards

JHoelli:
# New Workflow for Dashboards 

## New Dialog Component
Combining all Metadata: Permission, Titel, Assets
image

## Create New Dashboard

Metadata Dialog is triggered on create.
image

## Manage Dashboard
image
image

## Edit Dashboard
image

Former Properties from Create were moved to an extra panel. Introduced a live preview for those items.
image

## Tests 
Adapted Tests for new UI.
**NOTE that tests of resources not yet adapted to the new workflow dialog might fail - e.g., charts assets/persmissions, connect assests/Permissions, pipeline assets/permissions**


tenthe:
@JHoelli, thanks for opening the PR.

Could you please adjust the workflow for **Create New Dashboard**? When a user clicks the **+ New Dashboard** button, they should be redirected directly to the dashboard editor, with a **Save** button in the top-left corner. This would align the dashboard creation flow with the workflow used for other resources and provide a more consistent user experience.

As I understand it, the current implementation opens a dialog to create the dashboard, and the user then needs to edit it afterwards. It would be preferable to let users configure and save the dashboard directly from the editor.

ui/deployment/i18n/de.json
```diff
@@ -608,9 +609,12 @@
   "Mail Settings": "E-Mail-Einstellungen",
   "Mail Template": "Mail-Vorlage",
   "Make available to all users with appropriate role": "Allen Benutzern mit entsprechender Rolle zur Verfügung stellen",
+  "Manage": "Verwalten Sie",
```
tenthe:
```suggestion
  "Manage": "Verwalten",
```

tenthe (approved):

PR #4621, opened 2026-06-23
fix(#4618): Health check restarts adapter during stopping process

### LINKED ISSUE #4618: Health check restarts adapter during stopping process
### Body

If an adapter health check gets triggered to the same time when an adapter currently gets stopped, it can happen that the adapter gets restarted. To prevent this the health check should ignore adapters where it is starting or stopping

### StreamPipes Committer

SvenO3:
### Purpose

This prevents that the health check restarts adapters during start and stop transitions.
Now `AdapterTransitionRegistry` stores the transition state of adapters and the health check ignores these

### Remarks


streampipes-extensions-management/src/main/java/org/apache/streampipes/extensions/management/connect/AdapterTransitionRegistry.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+package org.apache.streampipes.extensions.management.connect;
+
+import org.apache.streampipes.model.health.AdapterInstanceState;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class AdapterTransitionRegistry {
```
tenthe, line 41: Could you add a brief description explaining the purpose of this service, why it was introduced, and what problem it solves? Having that context will make the intent clearer for future readers and help prevent it from being accidentally removed during future refactoring or cleanup efforts.

tenthe (approved):

PR #4632, opened 2026-06-24
fix(#4567): Prevent excessive dataset count queries when using large page sizes

### LINKED ISSUE #4567: Prevent excessive dataset count queries when using large page sizes
### Body

Prevent excessive dataset count queries when using large page sizes

The maximum table page size was recently increased to 500. While this works well for most overview pages, it can cause performance issues in the dataset overview.

The dataset overview automatically loads the **# Events (7d)** count for all datasets visible on the current page. When a page size of 500 is selected, opening the dataset overview can therefore trigger up to 500 count queries against the time-series storage.

Currently, the dataset overview uses the globally configured table page size and automatically requests event counts for all visible datasets. This can create unnecessary load on the backend and the underlying storage system, especially in environments with many datasets.

### Expected Behavior

Users should be able to browse large numbers of datasets without automatically triggering a large number of count queries.

### Possible Solution

One possible solution would be to decouple automatic count loading from the selected page size.

Potential approaches include:

* Limiting automatic count loading to a smaller number of datasets (e.g. 25 or 50)
* Loading counts on demand, like the option of **# Events (total)**
* Users can manually load the number of events by clicking on the new button

### Discussion

The proposed solution is only one possible approach. We should discuss whether there is a better solution to address the issue.

Alternative approaches could include backend-side batching, execution limits, caching of count results, or other mechanisms to reduce the load caused by count requests while maintaining a good user experience.

### StreamPipes Committer

SvenO3:
### Purpose

This fixes the excessive dataset count queries for large page sizes. `# Events (7d)` now only loads automatically for a maximum of 50 measurements und otherwise needs to be triggered manually similar to `# Events (total)`

### Remarks

### DIFF
--- a/ui/src/app/dataset/components/datalake-configuration/datalake-configuration.component.html
+++ b/ui/src/app/dataset/components/datalake-configuration/datalake-configuration.component.html
@@ -107,15 +107,41 @@
                                 >{{ 'Loading' | translate }}
                             </mat-spinner>
                         } @else {
-                            <sp-label
-                                tone="neutral"
-                                minWidth="100px"
-                                [labelText]="
-                                    configurationEntry.eventsLatest | number
-                                "
-                                size="small"
-                            >
-                            </sp-label>
+                            @if (configurationEntry.eventsLatest > -1) {
+                                <sp-label
+                                    tone="neutral"
+                                    class="cursor-pointer"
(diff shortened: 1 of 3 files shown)


SvenO3:
Not needed anymore. See #4633

PR #4637, opened 2026-06-25
feat(#4633): Move dataset count display into dataset overview

### LINKED ISSUE #4567: Prevent excessive dataset count queries when using large page sizes

### LINKED ISSUE #4603: Retrieving data lake measure counts for a large number of datasets
When trying to load the event count in the `Datasets` view for a large number of available measures, the API returns an error since the measurement names are provided as query parameters along with a GET request.

We can change this to a HTTP POST request with a body payload to avoid too large URL request parameters.

### LINKED ISSUE #4633: Move dataset count display into dataset overview
### Body

The querying of the dataset count courses multiple issues for a large number of datasets (see #4567, #4603).
To fix this a count should only be executed for one dataset at a time:
- The view of all dataset should therefore only show the time of the latest event instead of the the total number of events.
- The dataset overview should be expanded with a metrics tab that show the total number of events as well as a chart for the number of events over the last 7 days

### StreamPipes Committer

SvenO3:
### Purpose

This fixes #4567, #4603 by replacing the `# Events (7d)` and `# Events (total)`columns in the dataset overview with a `Last event` column that show the time of the last event. The event count is now only shown in a new metrics tab in the dataset details view that also shows a bar chart for the last 7 days.

This introduces two new REST endpoints for datalakes:
- `{id}/count`: Queries the count of a single measurement instead of a full set of measurements
- `/measurements/latest-events`: Queries the latest event time for a set of measurements

image

image

### Remarks

The `/api/v4/datalake/measure/count` endpoint is now deprecated


streampipes-rest/src/main/java/org/apache/streampipes/rest/impl/datalake/DataLakeMeasureResource.java
```diff
@@ -80,29 +79,53 @@ public ResponseEntity<?> addDataLake(@RequestBody DataLakeMeasure dataLakeMeasur
...
+        .getOrDefault(measure.getMeasureName(), 0));
+  }
+
+  @Operation(
+      summary = "Deprecated measurement count endpoint",
+      description = "Use /api/v4/datalake/measure/{id}/count instead.",
+      deprecated = true
+  )
+  @GetMapping(path = "/count", produces = MediaType.APPLICATION_JSON_VALUE)
+  @PreAuthorize("this.hasReadAuthority()")
```
tenthe, line 121: Do we need to keep this endpoint? If it is no longer required, could we mark it removal in a future release?

tenthe (changes requested):

tenthe (approved):

PR #4648, opened 2026-06-28
fix: PLC4X cache recovery after failed read execution

dominikriemer:
### Purpose

Avoid evicting the shared cached PLC connection from the adapter reader on every read failure. Instead, let the leased connection invalidate itself when PLC4X read execution fails, including synchronous failures from terminated transaction executors.

This should prevent one adapter from closing a shared PLC connection while other adapters for the same PLC are still using or recreating it, improving recovery after PLC disconnects.

### Remarks


tenthe (approved):

chrisdutz:
Might be worth having a look at the current state of 1.0.0-SNAPSHOT as A LOT has changed there.

PR #4657, opened 2026-06-29
fix: Reduce logging for time series storage and adapter preprocessing

dominikriemer:
### Remarks


tenthe:
Hi @dominikriemer,

I like the idea of reducing the logging noise.

While looking at the implementation, I wondered whether the `loggedConversionErrors` set could grow unbounded over time. Since the key includes the raw value, a long-running adapter with many distinct invalid values would add a new entry for each one.

Do you think we need to guard against this, for example by keying only by adapter/field/datatype or by using a bounded cache?

tenthe (approved):

PR #4684, opened 2026-07-03
#4683 fix(msteams): migrate sink to Power Automate workflow format

### LINKED ISSUE #4683: MS Teams sink no longer works: legacy O365 connectors deprecated
Microsoft has deprecated Office 365 Connectors / Incoming Webhooks (*.webhook.office.com), which the MS Teams sink relies on. As a result, existing webhooks fail and notifications no longer reach Teams.
Two issues with the current implementation:

The simple message format sends the legacy {"text": "..."} MessageCard payload, which the replacement Power Automate workflow endpoints don't accept (returns 202 but renders nothing, or 400).
A JSON escaping bug: message content is inserted into the payload via raw string formatting, so text containing quotes, backslashes, or newlines produces malformed JSON and an HTTP 400.

The sink needs to be migrated to the Power Automate workflow format (Adaptive Card envelope) with proper JSON serialization.

### StreamPipes Committer

obermeier:
### Purpose

Microsoft has deprecated Office 365 Connectors / Incoming Webhooks (*.webhook.office.com), which the MS Teams sink relies on. As a result, existing webhooks fail and notifications no longer reach Teams.
Two issues with the current implementation:
Fixes #4683

### DIFF
--- a/streampipes-extensions/streampipes-sinks-notifications-jvm/src/main/java/org/apache/streampipes/sinks/notifications/jvm/msteams/MSTeamsSink.java
+++ b/streampipes-extensions/streampipes-sinks-notifications-jvm/src/main/java/org/apache/streampipes/sinks/notifications/jvm/msteams/MSTeamsSink.java
@@ -40,285 +40,309 @@ import org.apache.streampipes.wrapper.standalone.StreamPipesNotificationSink;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import org.apache.http.Header;
 import org.apache.http.HttpHost;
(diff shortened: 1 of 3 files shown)


tenthe:
Hi @obermeier, thanks a lot for the PR! 
I was wondering what will happen to existing pipeline elements. Do we need a migration for them, or did the configuration inputs remain unchanged so that no migration is required? How will pipelines using the old implementation be handled?

obermeier:
Hi @tenthe, thanks for this important question!

Currently running pipeline elements should throw a 403 (at least some weeks ago), since the webhook endpoint is not accepting requests anymore.

The user should update the URL. Everything else should stay the same from a user perspective. In the implementation, the new schema is used, and the more complex URLs are accepted.

I think a migration to the new system in SP-Sink is not possible, since it requires the generation of a new webhook URL in the MS Teams-App.

tenthe (approved):

PR #4712, opened 2026-07-11
fix(#4382): Improve handling of existing tables in PostgreSQL sink

### LINKED ISSUE #4382: Improve Handling of Existing Tables in PostgreSQL Sink

Backend

### What happened?

Currently, the PostgreSQL Sink assumes full control over table creation. When a pipeline is started, the sink always attempts to execute a `CREATE TABLE` statement based on the incoming event schema. If the user configures the sink to write into a pre-existing PostgreSQL table, the pipeline can fail.

Specifically, two main errors can occur with existing tables:

1. If the table uses standard PostgreSQL types, like `DOUBLE PRECISION` or `CHARACTER VARYING`, the client crashes with a `No enum constant` error.

Image

2. If the table is missing a column, the sink throws a generic `Table does not match the EventProperties` error instead of explaining exactly what is missing or wrong.

Image

Supporting existing schemas without altering them is essential for production environments and enterprise data governance.

### How to reproduce?

To trigger the **first error**:

1. Set up a PostgreSQL database.
2. Create a table using types like `DOUBLE PRECISION` or `VARCHAR`. For example: 
```sql
CREATE TABLE sensor_data (
   "timestamp" BIGINT,
   "sensorId" VARCHAR(255),
   "mass_flow" DOUBLE PRECISION
);
```
3. Create an adapter with a data stream that matches this schema.
4. Create a new pipeline. Connect the data stream to the PostgreSQL sink.
5. Configure the sink with your database credentials.
6. Save and try to start the pipeline.

The pipeline will crash with the `No enum constant` error.

To trigger the **second error**:

1. Do the same as before, but use a table with missing columns. For example:
```sql
CREATE TABLE sensor_data (
   "timestamp" BIGINT
);
```
2. Save and try to start the pipeline again.

The pipeline will fail with the generic `Table does not match the EventProperties` error.

### Expected behavior

The PostgreSQL sink should be able to reliably write to existing PostgreSQL tables without trying to recreate them. Specifically, the sink needs to:

- Check if the target table already exists in the database
- Skip the `CREATE TABLE` statement if the table is already there
- Validate the incoming stream data against the existing database schema
- Show a clear error message in the UI if things do not match (e.g., exactly which column is missing or has the wrong data type)
- Seamlessly insert the data if everything is correct, including support for batch inserts and timestamps

To achieve this behavior, we can make the following changes in the codebase:

- Use `information_schema.columns` to securely read the existing table schema from PostgreSQL
- Fix the `DbDataTypes` enum and the JDBC mapping logic so that it correctly translates multi-word PostgreSQL types like `DOUBLE PRECISION` or `CHARACTER VARYING`
- Bump the pipeline element version and define a migration in the Extension Service. If we add a new UI toggle (like "Append to existing table"), this step guarantees backward compatibility. It ensures that older user pipelines don't break and are automatically updated to the new format on startup

_No response_

stefaniesinner:
### Purpose

This continues #4383 and fixes #4382 by improving the PostgreSQL sink in three ways:

- Adds an append mode for existing tables and batch inserts
- Only checks the table in the current schema, so same-name tables in other schemas no longer break validation
- Lists all missing columns, or all type mismatches with readable type names, in one error message

A v0 to v1 migration keeps existing pipelines unchanged. 

I added tests (validation, migration, batching) and updated the PostgreSQL documentation with the new options.

### Results
Configuring the sink now shows a toggle for writing into an existing table and a batch size field. When enabled, the sink writes into the existing table instead of creating one, and events are inserted in batches of the given size.

Screenshot 2026-07-11 181216

### Remarks

### DIFF
--- a/streampipes-extensions/streampipes-sinks-databases-jvm/src/main/java/org/apache/streampipes/sinks/databases/jvm/jdbcclient/JdbcClient.java
+++ b/streampipes-extensions/streampipes-sinks-databases-jvm/src/main/java/org/apache/streampipes/sinks/databases/jvm/jdbcclient/JdbcClient.java
@@ -51,11 +51,10 @@ public class JdbcClient {
   protected StatementHandler statementHandler;
 
   /*
-   * Controls what happens when the target table does not exist yet.
-   * If false, the sink creates a new table. Otherwise, it writes into a
-   * table provided by the user and fails if it is missing.
+   * Controls whether the sink may create the target table if it does not exist yet.
+   * If false, the sink only writes into a table provided by the user and fails if it is missing.
    */
-  protected boolean appendToExisting = false;
+  protected boolean allowNewTableCreation = true;
 
   /*
    * Number of events to buffer before one batch insert (one round trip to the database).
@@ -196,12 +195,12 @@ public class JdbcClient {
       rs.close();
       if (tableAlreadyExists) {
         validateTable();
-      } else if (this.appendToExisting) {
-        // The user wants to use an existing table, so the sink does not create a new one
+      } else if (!this.allowNewTableCreation) {
+        // The user does not allow the sink to create tables, so the table has to be there already
         throw new SpRuntimeException("Table '" + this.tableDescription.getName()
-                + "' does not exist, but the option 'Use existing table' is enabled. "
-                + "Check that the table name matches an existing table in your database, "
-                + "or disable the option to let the sink create one automatically.");
+            + "' does not exist and the option 'Allow New Table Creation' is disabled. "
+            + "Check that the table name matches an existing table in your database, "
+            + "or enable the option to let the sink create one automatically.");
       } else {
         createTable();
       }
@@ -233,7 +232,7 @@ public class JdbcClient {
     checkConnected();
     Map<String, Object> eventMap = event.getRaw();
     if (!this.tableDescription.tableExists()) {
-      if (this.appendToExisting) {
+      if (!this.allowNewTableCreation) {
         throw new SpRuntimeException("Table '" + this.tableDescription.getName() + "' is not available.");
       }
       // Creates the table
@@ -245,23 +244,23 @@ public class JdbcClient {
       if (this.batchSize > 1) {
         // Collects events and sends them together in one batch insert
         this.statementHandler.addToBatch(
-                this.dbDescription, this.tableDescription,
-                connection, eventMap);
+            this.dbDescription, this.tableDescription,
+            connection, eventMap);
         if (this.statementHandler.getPendingBatchCount() >= this.batchSize) {
           this.statementHandler.executeBatch();
         }
       } else {
         // Sends one statement per event
         this.statementHandler.executePreparedStatement(
-                this.dbDescription, this.tableDescription,
-                connection, eventMap);
+            this.dbDescription, this.tableDescription,
+            connection, eventMap);
       }
     } catch (SQLException e) {
       boolean tableMissing = e.getSQLState() != null && e.getSQLState().startsWith("42");
-      if (tableMissing && !this.appendToExisting && this.batchSize <= 1) {
+      if (tableMissing && this.allowNewTableCreation && this.batchSize <= 1) {
         // If the table does not exists (because it got deleted or something, will cause the error
         // code "42") we will try to create a new one. Otherwise we do not handle the exception.
-        // Skipped when the user wants to use an existing table, or when events are still waiting to be written.
+        // Skipped when the user does not allow table creation, or when events are still waiting to be written.
         LOG.warn("Table '" + this.tableDescription.getName() + "' was unexpectedly not found and gets recreated.");
         this.tableDescription.setTableMissing();
         createTable();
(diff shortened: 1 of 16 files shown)


stefaniesinner:
The UI shows the new toggle's description twice. Also, it would probably be better to move the two new fields (e.g. below the table name) instead of leaving them at the bottom. I left both out here, since they are separate UI concerns.

tenthe:
> The UI shows the new toggle's description twice. Also, it would probably be better to move the two new fields (e.g. below the table name) instead of leaving them at the bottom. I left both out here, since they are separate UI concerns.

The field order is defined by the order of the properties in `declareConfig()`: the builder appends them in that order, and the UI renders `staticProperties` as-is. So if we want to change the order of these new fields, it should be done in this PR.

The duplicated text is a generic slide-toggle rendering issue. `sp-app-static-property` already shows the description, and `StaticSlideToggleComponent` renders the same `staticProperty.description` again inside the `mat-slide-toggle`. This should be addressed in a separate PR.

tenthe (approved):
@stefaniesinner thanks for the PR! Appreciate you taking care of this.

streampipes-extensions/streampipes-sinks-databases-jvm/src/main/java/org/apache/streampipes/sinks/databases/jvm/jdbcclient/JdbcClient.java
```diff
@@ -174,13 +187,20 @@ protected void ensureTableExists(String url, String databaseName) throws SpRunti
           this.dbDescription.getPassword());
       this.statementHandler.setStatement(connection.createStatement());
       ResultSet rs = connection.getMetaData().getTables(null, null, this.tableDescription.getName(), null);
-      if (rs.next()) {
+      boolean tableAlreadyExists = rs.next();
+      rs.close();
+      if (tableAlreadyExists) {
```
SvenO3, line 196: Currently the PostreSqlSink writes into an existing table even if `Use Existing Table` is disabled. Should this be possible or be prevented by throwing an error?
If the purpose is just to prevent the creation of new tables, we should probably rename it to something like `Require Existing Table` to indicate that it still can use an existing table even if the option isn't selected.

SvenO3 (commented):
Hi @stefaniesinner , thanks for your contribution. I had one question about the `Use Existing Table` option because at the moment events a written into an existing table even if this option is disabled?

I also encountered an error with empty PostreSQL schemas (see https://www.postgresql.org/docs/current/ddl-schemas.html) that caused an error when creating a new table:
```
ERROR: no schema has been selected to create in Position: 14
```
I created a quick fix for this (see https://github.com/apache/streampipes/commit/1ae6ce473177a685f4bf1030b223cad8fad7551a). This isn't really related to your changes but maybe you could also take a look at it

stefaniesinner:
> > The UI shows the new toggle's description twice. Also, it would probably be better to move the two new fields (e.g. below the table name) instead of leaving them at the bottom. I left both out here, since they are separate UI concerns.
> 
> The field order is defined by the order of the properties in `declareConfig()`: the builder appends them in that order, and the UI renders `staticProperties` as-is. So if we want to change the order of these new fields, it should be done in this PR.
> 
> The duplicated text is a generic slide-toggle rendering issue. `sp-app-static-property` already shows the description, and `StaticSlideToggleComponent` renders the same `staticProperty.description` again inside the `mat-slide-toggle`. This should be addressed in a separate PR.

Hi @tenthe, thank you for your feedback and the explanation!

In this case, I would move `Use Existing Table` directly below the table name, since it's a statement about the table. I would leave `Batch Size` at the bottom because it's a performance setting and putting it in the middle would split up the connection settings. 

Does that work for you?

image

streampipes-extensions/streampipes-sinks-databases-jvm/src/main/java/org/apache/streampipes/sinks/databases/jvm/jdbcclient/JdbcClient.java
stefaniesinner: Due to the data governance rules users in enterprise setups have to follow, I'd tend to rename it to `Require Existing Table`, as you suggest. But failing with an error (when the toggle is off and the user provides a table) seems like a valid option too. @SvenO3 and @tenthe, what do you think?

stefaniesinner:
@SvenO3, thanks for your feedback and the quick fix!

streampipes-extensions/streampipes-sinks-databases-jvm/src/main/java/org/apache/streampipes/sinks/databases/jvm/jdbcclient/JdbcClient.java
SvenO3: I'm also not sure what the better design would be. If I understand it correctly currently the toggle only prevents a user from creating a new table when `Use existing table` is set.
Maybe the logic could also be inverted and renamed to `Allow new table creation` if this is the only functionality the toggle wants to achieve. Or is there another purpose that I'm missing?

stefaniesinner: Yes, that's the only purpose. I looked at it again and I went with your suggestion: `Allow New Table Creation`, on by default. Because throwing an error when the table already exists would probably break existing pipelines. And `Allow New Table Creation` seems clearer than `Require Existing Table`. Is this fine with you?

stefaniesinner:
I refactored the code, including the rename discussed above, and addressed Copilot's comment about the hardcoded labels: the migration now builds them with `Labels.from`, like other migrators do. I also refactored and extended the tests.

dominikriemer (approved):

dominikriemer:
Thank you @stefaniesinner !

PR #4724, opened 2026-07-16
feat: implement consistent keyboard shortcuts and sequence navigation

kush2439p:
### Purpose

Implements #4695 — a consistent keyboard shortcut concept across StreamPipes.

- **Global navigation:** `g` then a letter jumps to a module (`g p` → Pipelines,
  `g d` → Dashboards, `g c` → Adapters, `g e` → Charts, `g a` → Assets,
  `g y` → Datasets, `g s` → Settings, `g h` → Home). Driven by a new
  `shortcutKey` field on each module in `modules.yml`, generated into
  `NavMenuService` — no duplicated route mapping.
- **Global utilities:** `?` opens Help, `Ctrl+B` / `Alt+B` toggles the sidebar.
- **Contextual shortcuts** (`E` to edit, `Ctrl+S` to save, etc.) unchanged.
- `KeyboardShortcutService` extended to support two-key sequences (1000ms
  timeout) alongside existing single-key/modifier shortcuts, plus new `alt`
  modifier support.
- Help page's Shortcuts tab now lists the navigation shortcuts dynamically
  from `NavMenuService`, so it stays in sync with active modules.
- Added unit tests: sequence fire/timeout, sequence precedence over a
  same-key contextual shortcut, alt-combo match, input-focus suppression.

### How to test

1. `npm install && npm run build-libraries && npm run build-dev`
2. Outside an input, press `g` then `p`/`d`/`c`/`e`/`a`/`y`/`s`/`h` → navigates.
3. Press `g`, wait >1s, then a letter → no navigation (timeout works).
4. Press `?` → Help opens; shortcuts listed there.
5. Press `Ctrl+B` / `Alt+B` → sidebar toggles.
6. `npm test` → `keyboard-shortcut.service.spec.ts` (6/6 passing).

### Remarks

While a `g`-sequence is pending, its second key always takes precedence over
a same-key contextual shortcut on that view (e.g. `g` `e` navigates to Charts
rather than triggering local edit mode). Intentional tradeoff for simplicity.

PR introduces (a) breaking change(s): no

PR introduces (a) deprecation(s): no

PR #4784, opened 2026-07-28
Shift key navigation prototype

tenthe:
## Overview

This PR builds on #4724. Thank you, @kush2439p, for providing the original implementation.

This follow-up explores an alternative, more discoverable interaction for global navigation. Instead of requiring users to learn the `g` sequences from the Help page, holding Shift reveals shortcut badges directly beside the available navigation items.

This is intended as a prototype and a starting point for discussing consistent shortcut behavior across StreamPipes, rather than as a final design. In particular, we still need to determine how existing contextual shortcuts should be incorporated into a consistent and discoverable concept.

## Changes

- Replaces `g` navigation sequences with `Shift+letter` navigation.
- Shows key badges beside permitted navigation items after holding Shift for 200 ms.
- Allows experienced users to navigate immediately without waiting for the badges.
- Keeps the collapsed sidebar compact and temporarily reveals collapsed navigation groups while hints are visible.
- Suppresses Shift navigation in editable fields and while dialogs are open.
- Uses the following mappings:
  - `Shift+H` → Home
  - `Shift+P` → Pipelines
  - `Shift+D` → Dashboards
  - `Shift+C` → Charts
  - `Shift+A` → Adapters
  - `Shift+X` → Assets
  - `Shift+T` → Datasets
  - `Shift+S` → Settings
- Changes `Shift+?` to open the existing shortcut overview in a temporary dialog.
- Keeps `Ctrl/Cmd+B` and `Alt+B` for toggling the sidebar.
- Continues deriving navigation shortcuts from `modules.yml` through `NavMenuService`.
- Adds tests for hint timing, editable-field suppression, and immediate Shift shortcuts.

grafik

grafik

## Remarks

The prototype currently applies key tips only to the main navigation. A future iteration could extend the same interaction to contextual actions and dialog buttons.

The main goal of this PR is to start a discussion about a shortcut system that is consistent, discoverable, accessible, and predictable across StreamPipes. The mappings, activation behavior, and visual presentation should all be considered part of that discussion.

@kush2439p, what do you think about these changes? Is there anything we could improve?

PR introduces breaking changes: no

PR introduces deprecations: no


kush2439p:
Thanks for building on my implementation! I really like the Shift-based hints because they make the shortcuts much easier to discover than memorizing the g sequences. Deriving the navigation shortcuts from modules.yml also looks clean. I’ll test the prototype, especially the collapsed sidebar and input/dialog behavior. One thing worth checking is that the new shortcut descriptions support translations and that shortcut keys remain unique when modules change. Overall, I think this is a good direction.

tenthe:
@kush2439p thanks a lot for the feedback and for providing the original implementation! I’m glad you like the Shift-based hints and the module-derived navigation shortcuts. I’ve now added translations as you suggested. I’ll close your PR and merge this one. I’ll open a separate issue to clarify and align the remaining shortcuts, as they are not yet fully consistent.

SvenO3 (approved):
I also really like these shortcuts and their visibility thanks to the hints.
When playing around with the shortcuts I noticed that the `Esc` shortcut sometimes closes panels that can't be reopened when editing charts or dashboards. But that's probably a different issue

PR #4833, opened 2026-08-14
refactor(#4734): Align backend internal dataset terminology

### LINKED ISSUE #4734: Align backend internal dataset terminology

JHoelli:
StreamPipes currently uses dataset, data lake measure and measurement series for overlapping concepts. The UI and documentation introduce Dataset as reusable collection of data. This naming should be consistent across the whole project. This PR resolves the naming for non-interface classes in the backend. 

PR introduces (a) deprecation(s): yes, im the streampipes client


SvenO3 (changes requested):
I've tested a migration from dev to this branch and noticed that after the migrations adapters with a persist pipeline can no longer be created.
The adapter still tries to create a pipeline with a data lake and appId `org.apache.streampipes.sinks.internal.jvm.datalake`.  This results in the following error:
```
java.lang.IllegalArgumentException
	at java.base/java.util.Optional.orElseThrow(Optional.java:407)
	at org.apache.streampipes.storage.couchdb.impl.pipeline.DataSinkStorageImpl.getFirstDataSinkByAppId(DataSinkStorageImpl.java:46)
	at org.apache.streampipes.storage.couchdb.impl.pipeline.PipelineElementDescriptionStorageImpl.getDataSinkByAppId(PipelineElementDescriptionStorageImpl.java:81)
	at org.apache.streampipes.manager.pipeline.compact.generation.PipelineElementConfigurationStep.makeSink(PipelineElementConfigurationStep.java:68)
	at org.apache.streampipes.manager.pipeline.compact.generation.PipelineElementConfigurationStep.lambda$apply$0(PipelineElementConfigurationStep.java:50)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1612)
	at org.apache.streampipes.manager.pipeline.compact.generation.PipelineElementConfigurationStep.apply(PipelineElementConfigurationStep.java:44)
	at org.apache.streampipes.manager.pipeline.compact.CompactPipelineManagement.makePipeline(CompactPipelineManagement.java:48)
```

SvenO3 (approved):

dominikriemer:
I have a few comments we need to address before merging:

* MigrateDataLakeDatabaseToDatasetMigration loads all documents into a single array. There is a high chance that this HTTP request fails due to high response times or document size. It's probably better to migrate one-by-one (with checkpoints) or to leave the database name as is.
* Why does the DataLakeSink have a new version (3) even though there is no migration?
* `DataLakeMeasure` is marked as deprecated but moved into a new package, so this is more a breaking change than a deprecation -> we should decide if we completely replace it or leave the deprecated class in the same package
* Since we don't migrate the models in the UI, the backend code and typescript code can diverge (e.g. a wrong class name is assigned to the `DataLakeMeasure` in `streampipes-model.ts`, which becomes a `DatasetMetadata` class later in `streampipes-rest`. Does this cause any problems?

JHoelli:
- [ ] MigrateDataLakeDatabaseToDatasetMigration loads all documents into a single array. There is a high chance that this HTTP request fails due to high response times or document size. It's probably better to migrate one-by-one (with checkpoints) or to leave the database name as is.
- [X] Why does the DataLakeSink have a new version (3) even though there is no migration? 
    --> Legacy change. In a first version we migrated the data sink via data sink migration. Due to the highly intertwined data sink, the migration is now moved to the Core, to be able to also migrate all other related documents !
- [X] DataLakeMeasure is marked as deprecated but moved into a new package, so this is more a breaking change than a deprecation -> we should decide if we completely replace it or leave the deprecated class in the same package
 --> Added path to deprecation, only for DataLakeMeasure as this is the only part the deprecated client api relies on.
- [X] Since we don't migrate the models in the UI, the backend code and typescript code can diverge (e.g. a wrong class name is assigned to the DataLakeMeasure in streampipes-model.ts, which becomes a DatasetMetadata class later in streampipes-rest. Does this cause any problems?
--> now migrated, was working due to the deprecation of DataLakeMeasure, would break as soon as deprecated class is removed

PR #4842, opened 2026-08-23
Show the certificate subject in the certificate list

### LINKED ISSUE #4841: Certificate list shows Issuer instead of Subject

UI

## Description
In the OPC UA certificate management UI (Settings → Certificates), the overview list displays the issuing CA's `issuerDn` (e.g. `CN=My Intermediate CA,O=My Organization,C=DE`) rather than the certificate's own `subjectDn`.

## Problem
When multiple servers are signed by the same intermediate/root CA, they all show identical information in this list, making it impossible to distinguish individual server certificates at a glance - even though the detail view (opened per certificate) already has everything needed to tell them apart.

## Suggestion
Show `subjectDn` as the primary identifier in the list instead of `issuerDn`. Additionally showing the `URI` entry from `subjectAlternativeNames` (the OPC UA ApplicationUri) would help too, since that's the identity that actually matters for OPC UA connections - `subjectDn`'s Common Name is often arbitrary free text, while the ApplicationUri is what's configured in the connecting application.

Both fields are already present in the per-certificate detail view, so this would just be surfacing existing data one level higher, in the list itself.

### How to reproduce?

_No response_

### Expected behavior

_No response_

_No response_

None

arpitjain099:
Closes #4841.

`certificate-label.component.html` rendered `certificate.issuerDn`, so every certificate signed by the same intermediate or root CA showed the same string and there was no way to tell two servers apart without opening each one.

It now shows `subjectDn`, and the OPC UA ApplicationUri underneath it, taken from the `URI` entry of `subjectAlternativeNames` as you suggested, since that is what an operator actually matches a server on. The DNS badges below are untouched. `subjectDn` and `subjectAlternativeNames` are both already on the `Certificate` model (`streampipes-model.ts:721-738`), so nothing needed to change on the API side.

One thing outside the strict scope: `ngOnInit` had a leftover `console.log(this.dnsNames)` sitting in the lines I was editing, so it is gone too. Happy to put it back if you would rather keep this to the reported change.

On verification, so the claim is not larger than what I did: this is a template and component change and I have not run the UI. I checked the field names against the generated model and ran the repo's own prettier config (3.7.4, `ui/.prettierrc.json`) over both files, which reports "All matched files use Prettier code style". I have not run `ng lint` or built the frontend, so CI is the real check there.

### DIFF
--- a/ui/src/app/configuration/extensions-service-management/certificate-configuration/certificate-label/certificate-label.component.html
+++ b/ui/src/app/configuration/extensions-service-management/certificate-configuration/certificate-label/certificate-label.component.html
@@ -18,8 +18,13 @@
 
 <div fxLayout="column" class="pt-sm pb-sm">
     <span class="text-sm">
-        {{ certificate.issuerDn }}
+        {{ certificate.subjectDn }}
     </span>
+    @if (applicationUri) {
+        <span class="text-sm">
+            {{ applicationUri }}
+        </span>
+    }
     <div fxLayout="row wrap" class="mt-sm">
         @for (dnsName of dnsNames; track $index) {
             <sp-label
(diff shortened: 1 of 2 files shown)


dominikriemer (approved):

dominikriemer:
Thank you @arpitjain099 !

PR #4878, opened 2026-09-04
fix(connect): support configurable decimal separator in CSV parser

### LINKED ISSUE #530: CsvFormat Decimal Support
If I create a FileSet with a CSV file that uses "," as decimal seperator instead of "." the `CsvFormat` does not recognize those values as numeric values and treats them as a string.

Imported from Jira [STREAMPIPES-353](https://issues.apache.org/jira/browse/STREAMPIPES-353). Original Jira may contain additional context.
Reported by: bossenti.

manish9363:
Closes #530

CSV values using `,` as the decimal separator were detected as strings instead of numbers. This adds a "Decimal separator" option to the CSV parser and makes `DatatypeUtils` normalize the configured separator before numeric detection.

Behavior is additive and backward-compatible: existing `DatatypeUtils` overloads default to `.`, and normalization is skipped for ambiguous values (e.g. `1,000,000`) to avoid corrupting grouped numbers.

**Tested:** new unit tests in `DatatypeUtilsTest` (comma decimals + guard cases) and `CsvParserTest` (`;` delimiter with `,` decimals). `mvn test` green for both affected modules with 0 checkstyle violations.

**Open question for maintainers:** I implemented lenient behavior (configuring `,` still also accepts `.`-values). Happy to switch to strict locale honoring if preferred.

### DIFF
--- a/streampipes-connect-shared/src/main/java/org/apache/streampipes/connect/shared/DatatypeUtils.java
+++ b/streampipes-connect-shared/src/main/java/org/apache/streampipes/connect/shared/DatatypeUtils.java
@@ -31,6 +31,12 @@ public class DatatypeUtils {
   private static final Logger LOG = LoggerFactory.getLogger(DatatypeUtils.class);
 
   /**
+   * The decimal separator assumed by {@link Double#parseDouble(String)} and
+   * {@link org.apache.commons.lang3.math.NumberUtils#isParsable(String)}.
+   */
+  public static final char DEFAULT_DECIMAL_SEPARATOR = '.';
+
+  /**
    * Converts the given value to a specified XSD datatype.
    * This method attempts to convert the input value to the target datatype specified by the XSD string.
    * It supports conversion to string, double, float, boolean, integer, and long types.
@@ -53,6 +59,27 @@ public class DatatypeUtils {
                                     Object value,
                                     String targetDatatypeXsd,
                                     AtomicBoolean loggedConversionError) {
+    return convertValue(adapterName, value, targetDatatypeXsd, DEFAULT_DECIMAL_SEPARATOR, loggedConversionError);
+  }
+
+  /**
+   * Converts the given value to a specified XSD datatype using the provided decimal separator.
+   * When the input is a string representing a floating point number that uses a decimal separator
+   * other than {@code '.'} (e.g. the {@code ','} common in many European locales), the separator
+   * is normalized before parsing so that the value is treated as numeric rather than as a string.
+   *
+   * @param adapterName The adapter whose event value should be converted.
+   * @param value The value to be converted.
+   * @param targetDatatypeXsd The target XSD datatype as a string.
+   * @param decimalSeparator The decimal separator used in string representations of numbers.
+   * @param loggedConversionError Guard that ensures a conversion error is logged at warn level only once.
+   * @return The converted value, or the original value if conversion fails.
+   */
+  public static Object convertValue(String adapterName,
+                                    Object value,
+                                    String targetDatatypeXsd,
+                                    char decimalSeparator,
+                                    AtomicBoolean loggedConversionError) {
     if (value == null) {
       return null;
     }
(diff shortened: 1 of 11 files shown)


tenthe:
Hi @manish9363,

thanks a lot for the PR!

Could you please confirm that you reproduced the issue in a running StreamPipes instance and tested the fix end-to-end locally?

It would also be great if you could add a few screenshots showing the issue before the fix and the working behavior after the fix. This helps us verify that the change was tested in a real running system.

manish9363:
> Hi @manish9363,
> 
> thanks a lot for the PR!
> 
> Could you please confirm that you reproduced the issue in a running StreamPipes instance and tested the fix end-to-end locally?
> 
> It would also be great if you could add a few screenshots showing the issue before the fix and the working behavior after the fix. This helps us verify that the change was tested in a real running system.

Yes, I reproduced the issue and tested the fix end-to-end in a running StreamPipes instance (Docker Compose). Screenshots below.

Setup: I used a CSV with ; as the column delimiter and , as the decimal separator:

sensor;temperature;humidity
sensor-a;3,14;45,5
sensor-b;21,7;60,2

Before (unpatched): In Connect → File Stream → CSV, temperature and humidity are detected as STRING, and the event preview keeps them as text ("temperature": "3,14").

After (with this fix): The CSV parser now exposes a "Decimal separator" option. Setting it to , makes temperature and humidity correctly detected as FLOAT, and the preview shows real numbers ("temperature": 3.14). sensor stays a STRING as expected.
before
after

manish9363:
@tenthe : Let me know if you need anything else

SvenO3 (changes requested):
@manish9363 Thank you for your contributions. The changes look good to me. Just 2 things that need to be fixed:
1. The Cypress tests that use the file stream adapter like for example `fileStream.spec.ts` need to add the new input field to run successfully
2. Values with a period `.` like 100.000 should not get parsed as float when a different decimal separator is selected. I think that's the better design because a period could also be used as a grouping separator for large numbers

manish9363:
> @manish9363 Thank you for your contributions. The changes look good to me. Just 2 things that need to be fixed:
> 
> 1. The Cypress tests that use the file stream adapter like for example `fileStream.spec.ts` need to add the new input field to run successfully
> 2. Values with a period `.` like 100.000 should not get parsed as float when a different decimal separator is selected. I think that's the better design because a period could also be used as a grouping separator for large numbers

@SvenO3 Thanks for the review. I have 
Screenshot 2026-09-07 at 11 20 49 AM
Fixed both points:

- Values with a different separator (e.g. 100.000 when "," is selected)
  now stay strings instead of being parsed as float.
- Updated the Cypress tests that use the file stream adapter to fill the
  new decimal separator field.

Verified locally in a running instance - 100.000 stays a string and 3,14
is parsed as a float.

tenthe:
Hi @manish9363,
thanks a lot for providing the update to the PR!

Did you notice that the following Cypress smoke tests are currently failing?

* `connect/persistInDataLake.smoke.spec.ts`
* `connect/rules/schemaRules.smoke.spec.ts`

Maybe you can also run the full Cypress test suite locally to make sure the new option does not break any existing tests?

If you need any help, just let us know.

manish9363:
Hi @tenthe
Thanks for the pointer! I ran the Cypress suite locally against a full build of the branch and reproduced both failures.

Root cause: the new decimal separator option makes the format section taller, which pushed the file-selected indicator into a scrollable area, so the `should('be.visible')` check timed out. I now scroll it into view before asserting (and did the same for the static property inputs).

Both `connect/persistInDataLake.smoke.spec.ts` and `connect/rules/schemaRules.smoke.spec.ts` pass locally now. I also ran the full smoke suite; the only remaining failures are the OPC-UA and Kafka specs (they need their broker/opc containers, which I didn't run locally) and are unrelated to this change.

Pushed as a5acc18.

SvenO3 (approved):

PR #4902, opened 2026-09-09
feat(#4889): Align permissions menu label and show element name in dialog

### LINKED ISSUE #4889: Align permissions menu label and show element name in dialog
### Body

Make the permissions menu entry in the data set view consistent with the other views, and indicate which element is being edited in the permissions dialog.

## Description

In the data set view, the menu entry is currently named "Manage permissions", while the corresponding entry in the other views is named "Manage". The data set entry should be changed to match the naming used elsewhere.

Image

In addition, when the permissions of a data set are edited, the dialog does not give any hint about the name of the element whose permissions are being edited.

Image

### StreamPipes Committer

SvenO3:
### Purpose

This PR improves dataset management by aligning the overview action with other managed resources by:
- Renaming the dataset action from “Manage permissions” to “Manage.”
- Displaying the selected dataset name in a read-only field within the management dialog.
- Enabling adding and removing dataset asset links.

### Remarks


tenthe:
Hi @SvenO3,
thanks for providing the PR.

Unfortunately, we can't rename a dataset. While I'd like to have this feature, we would need to separate the measurement name in InfluxDB from the dataset name in order to support it, which is a bigger change.

Until we have that, I'm in favor of just showing the name here and disabling the edit. What do you think, or do you have an alternative idea? Renaming the data set would brake existing pipelines and charts.

tenthe (approved):

PR #4918, opened 2026-09-11
feat (#4569): add asset hierarchy move action

### LINKED ISSUE #4569: Allow merging asset models in the assets overview
### Body

In the assets overview, users should be able to move an asset model into another asset model.

This operation should merge the two asset structures by making the selected asset model part of the target asset model. The goal is to simplify restructuring of existing asset hierarchies and support consolidation of asset models without requiring manual recreation.

### StreamPipes Committer

JHoelli:
- Adds a Move action and destination-tree dialog for top-level assets and editable subassets.

**In Asset Edit Mode: Option to move Subassets to a different Asser**
image
**In the Overview: Option to move the wohle Asset**
image
**Dialog Window**
image

- Transfers the complete selected hierarchy, including asset links and metadata.
- Warns about site mismatches and removes the moved asset’s site assignment when confirmed.


dominikriemer:
@JHoelli the dialog should have the same look and feel as the other dialogs, for actionable dialogs use SLIDE_IN which also applies our global toolbar colors.

JHoelli:
image

tenthe:
@JHoelli, right now a sub-asset can only be moved under another asset. It should also be possible to promote a sub-asset to a top-level asset model, i.e. detach it from its current parent and save it as a new root model with its own elementId.

JHoelli:
image

I am a bit unsure about the spot and the naming of the option. 

Other naming options would be : 
- Promote to root asset
- Promote to top-level asset 
- Make top-level asset
- Move to top level

The option could also be added to the move-asset-dialog, instead of the asset menu.

JHoelli:
image

tenthe (approved):

PR #4936, opened 2026-09-14
fix(#1614): ClassCastException in DataLakeSink

JHoelli:
image

- Detects schema/runtime property-type mismatches before primitive casting.
- Reports the first mismatch as a runtime error; later occurrences are debug-only until restart.
- Preserves valid fields after subsequent malformed events.
- Refreshes extension monitoring before returning pipeline logs.


dominikriemer:
I've updated malformed-field handling to report a monitoring warning once per runtime name without throwing. Also improved added a reporter for preserving valid fields from the first event onward instead of throwing an exception on the first event. Added a point.hasFields() guard to skip empty writes.

